PR #5, opened 2023-01-31
Producing .v files

pierrevial:


Cargo.toml
```diff
@@ -5,7 +5,13 @@ edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+# the profile [snapshot] should be customized so that snapshot files are automatically generated
+[profile.snapshot]
+inherits = "dev"
+
 [dependencies]
+cargo-insta = "1.26.0"
+insta = "1.26.0"
```
clarus, line 14: Without changes to this file if `insta` is not used.

src/main.rs
```diff
@@ -16,8 +16,11 @@ extern crate rustc_middle;
 extern crate rustc_session;
 extern crate rustc_span;
 
-use std::{path, process, str};
+extern crate insta;
```
clarus, line 19: Same, without insta if this is not used

src/main.rs
```diff
@@ -830,81 +833,107 @@ impl TopLevel {
...
+                n_tests += 1;
+                let snap_path = file_parent.to_str().unwrap().to_string() + "/" + file_stem.to_str().unwrap() + ".snapshot";
+                let mut snap_file = fs::File::open(snap_path).unwrap();
+                let mut snap_contents = String::new();
+                snap_file.read_to_string(&mut snap_contents).unwrap();
+                let mut file = fs::File::open(&path).unwrap();
+                let mut contents = String::new();
+                file.read_to_string(&mut contents).unwrap();
+                assert_eq!(contents,snap_contents,"The test failed on {}\n",file_stem.to_str().unwrap());
+                // Some dead code using unsucessfully the insta crate
```
clarus, line 929: Without dead code

clarus (commented):
There should be a way for the tests to be run into the CI (maybe that is already the case) and a comment in the README to know how to re-generate the snapshots.

PR #6, opened 2023-02-01
Indentation

dphov:
This PR holding changes for generating "readable" code with indentation


src/main.rs
```diff
@@ -16,18 +16,29 @@ extern crate rustc_middle;
...
 
 use pretty::RcDoc;
+
+mod render;
 use rustc_errors::registry;
 use rustc_session::config::{self, CheckCfg};
 use rustc_span::source_map;
 
+#[derive(Debug)]
+
```
clarus, line 29:
```suggestion
```

src/render.rs
```diff
@@ -0,0 +1,31 @@
+use pretty::RcDoc;
+
+pub(crate) fn paren(with_paren: bool, doc: RcDoc<()>) -> RcDoc<()> {
+    if with_paren {
+        RcDoc::text("(").append(doc).append(RcDoc::text(")"))
```
clarus, line 5: Maybe we can use https://docs.rs/pretty/latest/pretty/struct.RcDoc.html#method.concat to reduce the number of `append` and with `use pretty::RcDoc` avoid the need for the prefix `RcDoc`.

src/render.rs
```diff
@@ -0,0 +1,31 @@
...
+    } else {
+        doc
+    }
+}
+
+pub(crate) fn bracket(doc: RcDoc<()>) -> RcDoc<()> {
```
clarus, line 11: OK I did not know the `pub(crate)`.

dphov: `pub(crate)` will set _crate only_ visibility for a function.

src/render.rs
dphov: Resolved in https://github.com/formal-land/coq-of-rust/pull/6/commits/d2dc19a81f56ea548cf2b6ff72af83ba6ecd15c0 commit

PR #8, opened 2023-02-02
Compiling more items 1

pierrevial:


src/main.rs
```diff
@@ -130,6 +138,10 @@ enum TopLevelItem {
         name: String,
         body: TopLevel,
     },
+    Type {
```
clarus, line 141:
```suggestion
    TypeDefinition {
```
to differentiate it from the notion of `Type`

src/main.rs
```diff
@@ -118,6 +118,14 @@ enum Expr {
     },
 }
 
+/// [RustTypeKind] represents the different kinds of rust types: [struct]s,
+/// [enum]s, aliases...
+enum RustTypeKind {
```
clarus, line 123:
```suggestion
enum TypeDefinition {
```

src/main.rs
```diff
@@ -118,6 +118,14 @@ enum Expr {
...
+    RustStruct,
+    RustEnum,
+    RustAlias,
```
clarus, line 126:
```suggestion
    Struct,
    Enum,
    Alias,
```
as this is an AST in-between Coq and Rust.

clarus:
Closing to remove confusion.

PR #11, opened 2023-02-09
CLI and library interface

dphov:


dphov:
@clarus PR is ready for review

CLI translates rust file or folder and creates output folder
In case of source as folder, replicates inherited folder structure with Coq files in place of Rust files

`./coq_translation/[name_of_file | name_of_folder]/[some_translated_file.v]...`

Also, CI fails because there is no tests for now...

clarus (approved):

PR #13, opened 2023-02-13
fix: add pipeline to replace dollar in string, to make a valid coq name string

dphov:


lib/src/path.rs
```diff
@@ -45,8 +45,20 @@ pub fn compile_qpath(qpath: &rustc_hir::QPath) -> Path {
     }
 }
 
+/// WIP: should treat string to be a valid in Coq environment
```
clarus, line 48: Remove comment

PR #41, opened 2023-04-12
Added Rhs Support for Binary Operators

InfiniteEchoes:


.gitconfig
```diff
@@ -0,0 +1,7 @@
+[http]
```
clarus, line 1: You can ignore this file by adding it in `.git/info/exclude` https://luisdalmolin.dev/blog/ignoring-files-in-git-without-gitignore/ as I think this is specific to your use case.

coq_translation/examples-from-rust-book/CoqOfRust.v
```diff
@@ -396,94 +396,131 @@ Module std.
     End Add.
 
     Module Sub.
-      Class Trait {Output : Set} (Self : Set) : Set := {
+      Global Unset Primitive Projections.
```
clarus, line 399: OK top! I do not remember why I added this line, maybe it was only a test. If this still compiles, you can remove it:
```suggestion
```

PR #45, opened 2023-04-21
replace IndexedField with Notation.Dot

klausnat:
Make translated code compile


lib/src/pattern.rs
```diff
@@ -132,14 +132,16 @@ impl Pattern {
                 },
             ]),
             Pattern::TupleStruct(path, fields) => {
-                let signature_in_parentheses_doc = paren(
-                    true,
+                return nest([
+                    text("'"),
```
clarus, line 136: Redundant as already added elsewhere now. That should allow to add the file `functions/associated_functions_and_methods.v` to https://github.com/formal-land/coq-of-rust/blob/main/.github/workflows/rust.yml !

klausnat: indeed, fixed.

clarus:
Top, thanks!

PR #56, opened 2023-05-03
Add Imp to represent imperative source languages

dhilst:
The idea is to discuss to find a good representation of the imperative/impure constructs from our source languages (solidity and Rust) that are not easily or possibly mappable directly to gallina

### DIFF
--- a/coq_translation/Imp.v
+++ b/coq_translation/Imp.v
@@ -27,8 +27,12 @@ Inductive Ctl_ty : Set :=
 | Panic (** unrecovarable errors *)
 | Exception. (** recovarable errors/user errors *)
 
-(** [Mem i] is the memory indexed by the type i, it returns an existential type *)
-Parameter Mem : Set -> (forall a : Set, a).
+(** The type of a memory value which depends on an index *)
+Definition Mem_val_type {a : Set} (i : a) : Set. Admitted.
+(** A memory area indexed by the type [a] *)
+Definition Mem_area (a : Set) : forall (i : a), Mem_val_type i. Admitted.
+Check Mem_area nat. (* A memory area indexed by naturals *)
+Check Mem_val_type 0.
 
 (** [Imp] defines a free monad for a imperative language *)
 Inductive Imp : Set -> Set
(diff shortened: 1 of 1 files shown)


coq_translation/Imp.v
```diff
@@ -0,0 +1,48 @@
...
+| Break (** break loops *)
+| Continue (** restart loops *)
+| Panic (** unrecovarable errors *)
+| Exception. (** recovarable errors/user errors *)
+
+
+(** [Mem i a] is the memory indexed by the type i, containing the type a 
+    
+    I think this is not enough to represent memory, each "memory cell"
+    will have its own type, maybe a list of dependent pairs?
```
clarus, line 28: Yes, for that having an un-interpreted free-monad will be useful as it will allow us to have an operator to read the memory at an address with a certain expected type, without having to implement a cast mechanism to get a result of the correct type.

coq_translation/Imp.v
```diff
@@ -0,0 +1,48 @@
...
+*)
+Parameter Mem : Set -> Set -> Set.
+
+(** [Imp] defines a free monad for a imperative language *)
+Inductive Imp : Set -> Set
+  (** Set -> Set is a guess here to be to make it a mappable *)
+  := 
+| Pure {a : Set} : a -> Imp a (** pure expressions *)
+| Impure {a : Set} : a -> Imp a
+  (** impure expressions, like function calls that may fail,
```
clarus, line 39: Yes, this seems really similar to `Pure`

coq_translation/Imp.v
```diff
@@ -0,0 +1,48 @@
...
+      not sure if I need this *)
+| Loop {a b : Set} : list a -> Imp a (** may loop return values? *)
```
clarus, line 41: Yes, apparently it is possible to return from loops in Rust! https://doc.rust-lang.org/rust-by-example/flow_control/loop/return.html Maybe we do not need to support it though.

I was thinking about the following type for a `Loop`:
```coq
| Loop : Imp unit -> Imp unit
```
where it keeps evaluating its parameter until the answer is `Break`.

coq_translation/Imp.v
dhilst: I was in doubt about that, I may have expressions like `1`, `true` or even `[1,2,3]` which is known to be pure, but function calls `f()` may be impure both are one expression of type `a`. Maybe `Impure` should return `Imp (result a)`?

dhilst: I changed the type of `Impure` to `a -> Imp (result a e)`, I wonder if this `e` can be fixed as `string` without problem

coq_translation/Imp.v
dhilst: I changed the definition to `Parameter Mem : Set -> (forall a : Set, a).` Is this correct?

The `forall a : Set, a` is the dependent pair right? But this looks a lot like the `axiom : forall a : Set, a` which make me fells that I'm doing something wrong

coq_translation/Imp.v
dhilst: Fixed, I place your comment in the commentary regarding the constructor

coq_translation/Imp.v
dhilst: I removed the `WriteMem` and `ReadMem` constructors from `Imp` and try to get a better definition of memory, now I have
`Definition Mem_val_type {a : Set} (i : a) : Set. Admitted.` which is the type of a memory value which depends on an index `i`, and `Definition Mem_area (a : Set) : forall (i : a), Mem_val_type i. Admitted.` which is a memory area indexed by the type `a`. This seems better, what you think?

coq_translation/Imp.v
```diff
@@ -0,0 +1,72 @@
...
+      Pure a
+    | Free (f (Free a f))
+
+*)
+Module Free.
+  (* Inductive t (a : Set) (fnctr : Set -> Set) : Set := *)
+  (* | Pure : a -> t a fnctr *)
+  (* | Free : ??? *)
+  (* . *)
+End Free.
```
dhilst, line 68: I'm still struggling with the free monad definition and I think the Imp type is wrong as it is not parametrized by a functor, I'm leaving this comment and will work more on it tomorrow

PR #55, opened 2023-05-03
File Separation Experiment

InfiniteEchoes:
This PR has:
- Seperated the content in `CoqOrRust.v`. In particular,
  - `lib.v` contains the miscellaneous notation definitions
  - Other files are mostly the Coq support for Rust's `std` library.
  - Generated `Makefile` for building

What this PR hasn't done:
- Ensure the Coq code from Rust can utilize this new library.

PR #59, opened 2023-05-06
Updated files format according to comments

InfiniteEchoes:
This PR has:
1. Used a local repository directly from the original repository
2. Updated file contents according to comments mentioned in #55


clarus:
OK cool! I think we can merge, as this is a strict improvement to the existing implementation of the `CoqOfRust.v` library!

PR #61, opened 2023-05-10
Add docs/TRANSLATION.md

dhilst:


docs/TRANSLATION.md
```diff
@@ -0,0 +1,263 @@
...
+
+If  `[a]` is impure we need to use the monadic notation
+
+```
+  [let foo = a; ...]
+------------------------
+let** foo := [a] in [...]
+```
+* Note that `let*` will drive the type of `...`.
+* @TODO Can we know if `a` is impore or not?
```
clarus, line 145: We can assume that all the functions are impure to start with.

docs/TRANSLATION.md
```diff
@@ -0,0 +1,263 @@
...
+* @TODO `foo` is bounded in `...`. I had to take this into account
+  in coq-of-solidity because when I find the identifier `foo` inside
+  the `...` I need to know if it was local or not, to guide the translation.
+  Non local variables, i.e., state variables were translated to `state.(foo)`
+  while local variables stays like `foo`. In coq-of-solidity I use a map
+  as an environment to keep the scope of the variables (among other
+  things). I think that this is not a problem in rust because `self`
+  need to be explicited in order to be accessed.
```
clarus, line 153: Yes indeed `self` is like a normal variable.

docs/TRANSLATION.md
```diff
@@ -0,0 +1,263 @@
...
+[let Ctr foo = a else { ...0 ] }; ...1 ]
+----------------------------------------
+       let** foo := match [a] with 
+         | Ctr foo => [a] 
+         | _ => [ ...0 ]
+         end in
+         [ ...1 ]
+```
+
+* @TODO: Understand the `let**` notation, I'm just guessing that it
```
clarus, line 176: I added the `let**` as an optimization of the `let*` for some cases, but I am not sure if we should keep it. The two should be equivalent.

docs/TRANSLATION.md
```diff
@@ -0,0 +1,263 @@
...
+  * In coq-of-solidity I had to iterate over the expressions looking
+    for function calls, for other reason, but this can be used in the
+    translation to tell if an if expression is pure or not.
+
+## Function calls
+
+```
+  [f(a, b)]
+-------------
+ f(|[a] [b]|)
```
clarus, line 259:
```suggestion
 f(|[a], [b]|)
```

docs/TRANSLATION.md
dhilst: Okay I will use `let*` for now

dhilst:
Closing for now to focus on ERC20

PR #65, opened 2023-05-12
Add function application notation

dhilst:


lib/src/expression.rs
```diff
@@ -199,11 +198,69 @@ pub fn compile_expr(tcx: TyCtxt, expr: &rustc_hir::Expr) -> Expr {
...
+                            .map(|expr| match expr {
+                                Expr::Call { .. } => {
+                                    let vname = "x"; // needs to be a fresh name
+                                    fcalls_vars.push(vname);
+                                    Expr::Var(Path {
+                                        segments: vec![String::from(vname)],
+                                    })
+                                }
+                                e => e,
+                            })
```
dhilst, line 256: I have a question here: 

Context: I'm trying to transform `f(a,g(b))` into `let x = g(b); f(a, x)` (x needs to be not bounded but I'm ignoring this for now). At this point in the code, I'm compiling an expression, a function call, and I found other function calls in its arguments. `args` are those arguments. I need to replace the occurrences of function calls in `args` with references to fresh variable

Okay, my problem is, I want `args.iter().map(...).collect()` to have type Vec<Expr>`, this is okay for the `Expr::Call` branch because I create a new expression but for all other cases I just want to return the expression, them problem is that the `expr` has type  `&Expr` not `Expr`. What is the Rust way of doing this? I think I cannot go from `&Expr -> Expr` unless I copy it, and if it so, How do I copy it?

Another option is to return `&Expr::Var(...)` in the `Expr::Call` branch, then the whole line type checks, but now I need to return a new construct a new `Expr::Call` where args is `Vec<Expr>` and I have `Vec<&Expr>`

dhilst: I just have to use `into_iter` instead of `iter`

lib/src/path.rs
```diff
@@ -74,7 +74,7 @@ pub(crate) fn compile_qpath(qpath: &QPath) -> Path {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
dhilst, line 77: Not needed anymore 

```suggestion
#[derive(Debug)]
```

lib/src/pattern.rs
```diff
@@ -7,7 +7,7 @@ use rustc_middle::ty::TyCtxt;
 use rustc_span::source_map::Spanned;
 
 /// The enum [Pat] represents the patterns which can be matched
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
dhilst, line 10:
```suggestion
#[derive(Debug]
```

lib/src/expression.rs
```diff
@@ -173,6 +173,58 @@ fn tt() -> Expr {
...
+///  let y = g(x); f(a, y)
+fn monadic_translation(expr: Expr) -> Expr {
+    match expr {
+        Expr::Call { func, args } => {
+            let has_calls_or_lets = args.iter().any(|expr| {
+                (matches!(expr, Expr::Call { .. }) || matches!(expr, Expr::Let { .. }))
+            });
+            if !has_calls_or_lets {
+                return Expr::Call { func, args };
+            }
```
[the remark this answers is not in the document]
dhilst: okay, fixed with `match { ... if `

lib/src/expression.rs
```diff
@@ -173,18 +196,89 @@ fn tt() -> Expr {
...
+/// The above transformation may yield
+/// a let inside the function call, this is
+/// handled also, as below
+///
+///         f(let x = g(y); h(x))
+///  ----------------------------------
+///         let x = g(y); f(h(x))
+///  ----------------------------------
+///   let x = g(y); let z = h(x); f(z)
+fn monadic_translation(expr: Expr, fresh_vars: &mut FreshVars) -> Expr {
```
clarus, line 221: With a translation done as a second step after the first translation from the Rust AST to our AST.

lib/src/expression.rs
```diff
@@ -173,18 +196,89 @@ fn tt() -> Expr {
...
+                .into_iter()
+                .map(|expr| match expr {
+                    // Transform function calls into variables
+                    Expr::Call { .. } => {
+                        let vname = fresh_vars.next();
+                        let_vars.push((Pattern::Variable(vname.clone()), expr));
+                        Expr::Var(Path::local(vname))
+                    }
+                    // Handle the let inside arguments
+                    Expr::Let {
```
clarus, line 237: This should be handled as a particular case of handling all the expression cases in the monadic translation.

coq_translation/examples-others/monadic-transformation/example01.v
```diff
@@ -0,0 +1,25 @@
+(* Generated by coq-of-rust *)
+Require Import CoqOfRust.CoqOfRust.
+
+Import Root.std.prelude.rust_2015.
+
+Definition id (x : u64) : u64 := x.
```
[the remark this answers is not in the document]
clarus: The translation should also update all the other examples. There is the command to update all the snapshots in the README:
```sh
python run_tests.py
```

lib/src/expression.rs
```diff
@@ -173,18 +196,89 @@ fn tt() -> Expr {
...
+                .fold(fexpr, move |body, (pat, fcall)| Expr::Let {
+                    modifier: LetModifier::MonadicLet,
+                    pat,
+                    init: Box::new(fcall),
+                    body: Box::new(monadic_translation(body, fresh_vars)), // recurse in the body
+                });
+            let_expr
+        }
+        // Nothing to do, just return
+        e => e,
```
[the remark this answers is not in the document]
clarus: Yes, you should do the translation regardless if nested or not, for all the expressions.

lib/src/expression.rs
dhilst: Partially done, I created the boilerplate for the transformation but for now is just a big recursive identity. It runs inside `core.rs`, `create_translation_to_coq` 

```rust
    ...
    let result = rustc_interface::run_compiler(config, |compiler| {
        compiler.enter(|queries| {
            queries.global_ctxt().unwrap().enter(|tcx| {
                // Step 1 - Compile to top_level
                let top_level = compile_top_level(tcx);
                // Step 2 - Monadic Transformation
                let top_level = mt_top_level(top_level);
                // Step 3 - To string
                top_level.to_pretty(LINE_WIDTH)
            })
        })
    });
    ...
```

lib/src/top_level.rs
```diff
@@ -703,6 +708,154 @@ fn fn_to_doc<'a>(
...
+            args,
+            ret_ty,
+            body,
+            is_dead_code,
+        } => TopLevelItem::Definition {
+            name,
+            ty_params,
+            where_predicates,
+            args,
+            ret_ty, // @TODO do I need to transform the function type also?
```
clarus, line 779: Yes! Basically, the type arrow `A -> B` is translated to `A -> M B` instead. This can be done directly when we go from the Rust AST to the Coq AST.

lib/src/expression.rs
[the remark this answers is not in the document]
dhilst: These are still missing. The recursive rule it no needed? The remaining will be done in a second MR 
image

clarus:
For the pull-request description:

>     recursive function rules seem unneeded!?

Yes, it seems unneeded to me too. Maybe we will add some special primitive later to use the fuel for recursive functions.

>    I unnest let expressions, so let* a = (let* b = c in d ... becomes let* b = c in let a = d in ..., this may be the associativity law of monad

This can change the behavior of the code if the variable `b` is used later, in the `...`.

lib/src/top_level.rs
```diff
@@ -703,6 +703,167 @@ fn fn_to_doc<'a>(
...
+            name,
+            path,
+            is_glob,
+            is_type,
+        },
+        TopLevelItem::Error(err) => TopLevelItem::Error(err),
+    }
+}
+
+pub fn mt_top_level(tl: TopLevel) -> TopLevel {
```
clarus, line 863:
```suggestion
pub fn mt_top_level(top_level: TopLevel) -> TopLevel {
```

lib/src/top_level.rs
```diff
@@ -703,6 +703,167 @@ fn fn_to_doc<'a>(
...
+}
+
+fn mt_impl_items(items: Vec<(String, ImplItem)>) -> Vec<(String, ImplItem)> {
+    items.into_iter().map(mt_impl_item).collect()
+}
+
+fn mt_trait_item(body: (String, TraitItem)) -> (String, TraitItem) {
+    let (s, body) = body;
+    match body {
+        TraitItem::Definition { ty } => (s, TraitItem::Definition { ty }),
```
clarus, line 749: We will need to do a monadic transform for the types also after, to add the `M` at each `->`.

lib/src/expression.rs
```diff
@@ -173,6 +203,260 @@ fn tt() -> Expr {
...
+    fields
+        .into_iter()
+        .map(|field| mt_field(field, fresh_vars))
+        .collect()
+}
+
+pub fn mt_expressions(exprs: Vec<Expr>, fresh_vars: &mut FreshVars) -> Vec<Expr> {
+    exprs
+        .into_iter()
+        .map(|expr| mt_expression(expr, fresh_vars))
```
clarus, line 239: I think all these functions can be inlined in the code, as they are one-liners and used once (except for `mt_boxed_expression` that is probably used several times).

lib/src/expression.rs
```diff
@@ -173,6 +203,260 @@ fn tt() -> Expr {
...
+
+fn pure(e: Expr) -> Expr {
+    Expr::Pure(Box::new(e))
+}
+
+/// let a = (let b = c in d) in e
+/// ----------------------------------
+/// let b = c in
+/// let a = d in
+/// e
```
clarus, line 290: This can break if `b` is already used in `e` (maybe it is always fine thanks to the use of fresh variables)

lib/src/expression.rs
```diff
@@ -173,6 +203,260 @@ fn tt() -> Expr {
...
+/// let a' = MT(a);
+/// let b'; = MT(b);
+/// ...
+/// f'(a', b', ...)
+fn mt_call(func: Expr, args: Vec<Expr>, fresh_vars: &mut FreshVars) -> Expr {
+    let mut let_vars: Vec<(Pattern, Expr)> = vec![];
+    // Create one variable for the function
+    let fname = fresh_vars.next();
+    // Create one variable for each argument, take
+    // the oportunity to apply mt_expression into it
```
clarus, line 257:
```suggestion
    // the opportunity to apply [mt_expression] into it
```

lib/src/expression.rs
```diff
@@ -173,6 +203,260 @@ fn tt() -> Expr {
...
+    }
+}
+
+// @TODO add the translation logic (right now is just an ineficient identity)
+pub fn mt_expression(expr: Expr, fresh_vars: &mut FreshVars) -> Expr {
+    match expr {
+        Expr::Pure(x) => Expr::Pure(x),
+        Expr::LocalVar(x) => pure(Expr::LocalVar(x)),
+        Expr::Var(x) => pure(Expr::Var(x)),
+        // @TODO how to transform associated function?
```
clarus, line 340: I think it should be with `pure`, as they are a kind of identifier in the AST.

lib/src/expression.rs
```diff
@@ -173,6 +203,260 @@ fn tt() -> Expr {
...
+        Expr::AssociatedFunction { ty, func } => Expr::AssociatedFunction { ty, func },
+        Expr::Literal(x) => pure(Expr::Literal(x)),
+        Expr::AddrOf(box_expr) => pure(Expr::AddrOf(mt_boxed_expression(box_expr, fresh_vars))),
```
clarus, line 343: There should be a `let*` there

lib/src/expression.rs
```diff
@@ -173,6 +203,260 @@ fn tt() -> Expr {
...
+        Expr::Call { func, args } => mt_call(
+            mt_expression(*func, fresh_vars),
+            mt_expressions(args, fresh_vars),
+            fresh_vars,
+        ),
+        // @TODO I guess method call transformation should be similar to
+        // function application transformation
```
clarus, line 350: Yes

lib/src/expression.rs
```diff
@@ -173,6 +203,260 @@ fn tt() -> Expr {
...
+        },
+        Expr::Type { expr, ty } => Expr::Type {
+            expr: mt_boxed_expression(expr, fresh_vars),
+            ty,
+        },
+        Expr::Array { elements } => Expr::Array {
+            elements: mt_expressions(elements, fresh_vars),
+        },
+        Expr::Tuple { elements } => Expr::Tuple {
+            elements: mt_expressions(elements, fresh_vars),
```
clarus, line 393: There should be some `let*`

lib/src/expression.rs
```diff
@@ -173,6 +203,260 @@ fn tt() -> Expr {
...
+        Expr::Cast { expr, ty } => Expr::Cast {
+            expr: mt_boxed_expression(expr, fresh_vars),
+            ty,
+        },
+        Expr::Type { expr, ty } => Expr::Type {
+            expr: mt_boxed_expression(expr, fresh_vars),
+            ty,
+        },
+        Expr::Array { elements } => Expr::Array {
+            elements: mt_expressions(elements, fresh_vars),
```
clarus, line 390: With one `let*` per element

lib/src/expression.rs
```diff
@@ -173,6 +203,260 @@ fn tt() -> Expr {
...
+        Expr::Lambda { args, body } => pure(Expr::Lambda {
+            args,
+            body: mt_boxed_expression(body, fresh_vars),
+        }),
+        Expr::Seq { first, second } => Expr::Let {
+            modifier: "*",
+            pat: Pattern::Variable(String::from("_")),
+            init: mt_boxed_expression(first, fresh_vars),
+            body: {
+                fresh_vars.reset();
```
clarus, line 377: Can this be dangerous, as it can also change the value of the counter for the caller functions?

lib/src/expression.rs
dhilst: Question: It should be `let* x0 := (a, b, ...) in ...` or `let* x0 := a  in let* x1 := b in let* x2 := ... in (x0, x1, x2, ...)`

lib/src/expression.rs
dhilst: I didn't get the problem, I can remove the `.reset()` call, then it will reset at each definition

dhilst: Oh I get it, it will reset in the middle of a subexpression, you right, I'll remove it

lib/src/expression.rs
dhilst: you mean `#[inline]` or literally inlining them?

dhilst:
The building is failing in the monadic example 2
image

lib/src/expression.rs
clarus: Literally inlining

lib/src/expression.rs
dhilst: I did it, but I'm facing a problem with the generated code 

```
(* Generated by coq-of-rust *)
Require Import CoqOfRust.CoqOfRust.

Import Root.std.prelude.rust_2015.

Module Foo.
  Record t : Set := { _ : u32;}.
  
  Global Instance Get_0 : Notation.Dot 0 := {
    Notation.dot '(Build_t x0) := x0;
  }.
End Foo.
Definition Foo := Foo.t.

Module ImplFoo.
  Definition Self := Foo.
  
  Definition plus1 (s : Self) : M u32 :=
    let* α0 := (Pure s).[0] in
    let* α1 := Pure 1 in
    α0.["add"] α1.
  
  Global Instance AssociatedFunction_plus1 :
    Notation.DoubleColon Self "plus1" := {
    Notation.double_colon := plus1;
  }.
End ImplFoo.

(* #[allow(dead_code)] - function was ignored by the compiler *)
Definition main (_ : unit) : M unit :=
  let foo := Foo.Build_t (Pure 0) in
  let* _ := (Pure foo).["plus1"] in
  Pure tt.

```

I get this error 
```
Error:
The following term contains unresolved implicit arguments:
  (fun s : Self => let* α0 := (Pure s) .[ _] in (let* α1 := Pure 1 in (α0) .[ _] α1))
More precisely: 
- ?Dot: Cannot infer the implicit parameter Dot of Notation.dot whose type is "Notation.Dot 0" (no type class instance found) in environment:
  s : Self
- ?S: Cannot infer ?S in the partial instance "forall x : M Self, M ?S" found for the implicit parameter T of Notation.dot whose type is 
  "Set" in environment:
  s : Self
  x : M Self
- ?Dot0: Cannot infer the implicit parameter Dot of Notation.dot whose type is "Notation.Dot "add"" (no type class instance found) in
  environment:
  s : Self
  α0 : ?S@{x:=Pure s}
  α1 : Z
```

dhilst: I guess the instance that it is looking for is the `Global Instance Get_0` :thinking:

lib/src/top_level.rs
dhilst: I added `mt_ty` I need to think about the transformation on the types https://github.com/formal-land/coq-of-rust/pull/65/files#diff-3ce39e55f1f7c3b412ccdcff0a8c0af5892f535f1195ce14a7e1b43874a157d0R719

PR #67, opened 2023-05-14
Experiments around the state monad

clarus:


clarus:
I am merging what is there to close the pull-request. This provides a monad to allocate, read, and write pointers.

PR #72, opened 2023-05-29
Ink example 01

dhilst:


coq_translation/examples/traits/hash.v
```diff
@@ -28,7 +29,7 @@ Definition Person : Set := Person.t.
 Module Impl__crate_hash_Hash_for_Person.
   Definition Self := Person.
   
-  Definition hash (self : ref Self) (state : mut_ref __H) : M unit :=
+  Definition hash (__H : Set) `(_crate.hash.Hasher.Trait __H) (self : ref Self) (state : mut_ref __H) : M unit :=
```
dhilst, line 32: @clarus After implementing the `hash` module in `CoqOfRust.std.hash`, adding the instances, etc, and translating the hash.rs example again, The `Import` above and these two parameters `(__H : Set) ``(_crate.hash.Hasher.Trait __H)` were the only two changes that I had to do (by hand) in this file to make it typecheck.

dhilst:
Closing for now to focus on ERC20

PR #79, opened 2023-06-12
Add definitions for the Rust std library

clarus:


CoqOfRust/_CoqProject
```diff
@@ -0,0 +1,97 @@
+# !!!
+# Generated by configure.sh
```
InfiniteEchoes, line 2: For the CoqOfRust library part, since there are dependencies between the files, the order of the files could be necessary. I see that this file is generated, so I think it would be better if we can specify the order in some way, e.g. keep the original file or specify it in configure.sh

clarus: The ordering is automatically handled by the generated `Makefile`! I do not know how it works in detail, but there is a system from Coq itself to compute the dependency order of the files. The `Makefile` uses https://www.mankier.com/1/coqdep

Issue #80, opened 2023-06-13
Add ink examples


clarus:
Add the examples from https://github.com/paritytech/ink/tree/master/integration-tests into the project. A possible solution:

* have a git-submodule to copy their project
* have a shell script to compile their files and re-generate the examples files extracted from their code
* make sure that our configure script for examples try to compile them, unless if they are in the blacklist.

clarus:
For now, per https://github.com/formal-land/coq-of-rust/pull/83 , we only translate the `env` create of `ink!` https://github.com/paritytech/ink/tree/master/crates/env This is OK for now, and that would already be a huge achievement to translate this code to Coq code that type-checks.

PR #83, opened 2023-06-13
Add ink examples

clarus:
For now, only add the translation of https://github.com/paritytech/ink/tree/master/crates/env . All the Rust smart contracts from Ink! depend on the ink library, which itself depends on a lot of packages. We will need to see how far we go.


clarus:
Fix issue https://github.com/formal-land/coq-of-rust/issues/80

Issue #77, opened 2023-06-06
Use named type parameters for traits


clarus:
With traits, it is sometimes unclear what type parameter is referring to. Instead of using explicit ordered type parameters, we should use implicit named types parameters with curly braces.

clarus:
We can take the file https://github.com/formal-land/coq-of-rust/blob/main/examples/generics/generics_traits.rs as an example to translate with named parameters, as it contains a trait with type parameters:
```rust
// Non-copyable types.
struct Empty;
struct Null;

// A trait generic over `T`.
trait DoubleDrop<T> {
    // Define a method on the caller type which takes an
    // additional single parameter `T` and does nothing with it.
    fn double_drop(self, _: T);
}

// Implement `DoubleDrop<T>` for any generic parameter `T` and
// caller `U`.
impl<T, U> DoubleDrop<T> for U {
    // This method takes ownership of both passed arguments,
    // deallocating both.
    fn double_drop(self, _: T) {}
}

fn main() {
    let empty = Empty;
    let null = Null;

    // Deallocate `empty` and `null`.
    empty.double_drop(null);

    //empty;
    //null;
    // ^ TODO: Try uncommenting these lines.
}
```

Here is a translation that we can make:
```coq
(* Generated by coq-of-rust *)
Require Import CoqOfRust.CoqOfRust.

Import Root.std.prelude.rust_2015.

Module Empty.
  Inductive t : Set := Build.
End Empty.
Definition Empty := Empty.t.

Module Null.
  Inductive t : Set := Build.
End Null.
Definition Null := Null.t.

Module DoubleDrop.
  Class Trait {Self T : Set} : Set := {
    double_drop : Self -> T -> (M unit);
  }.
  
  Global Instance Method_double_drop `(Trait) : Notation.Dot "double_drop" := {
    Notation.dot := double_drop;
  }.
End DoubleDrop.

Module Impl_DoubleDrop_for_U.
Section Impl_DoubleDrop_for_U.
  Context {U T : Set}.

  Definition Self := U.
  
  Definition double_drop (self : Self) (Pattern : T) : M unit := Pure tt.
  
  Global Instance Method_double_drop : Notation.Dot "double_drop" := {
    Notation.dot := double_drop;
  }.
  
  Global Instance I : DoubleDrop.Trait (Self := U) (T := T) := {
    DoubleDrop.double_drop := double_drop;
  }.
End Impl_DoubleDrop_for_U.
[...]

PR #93, opened 2023-06-19
Named Type Parameters for Traits

InfiniteEchoes:
This PR is for solving the issues mentioned in #77 .


lib/src/top_level.rs
```diff
@@ -1459,14 +1459,21 @@ impl TopLevelItem {
...
+                              line(),
+                              text("Self"),
+                              line(),
+                              text(":"),
+                              line(),
+                              text("Set"),
+                              text(")"),
+                              line(),
+                              text("{"),
+                              concat(ty_params.iter().map(|ty| concat([text(ty), line()]))),
```
clarus, line 1472: I think this will not work if there is an empty list of `ty_params`.

lib/src/top_level.rs
```diff
@@ -1610,6 +1617,28 @@ impl TopLevelItem {
                             text("."),
                         ]),
                         hardline(),
+                        nest([
+                          text("Section"),
```
clarus, line 1621: The creation of a `Section`/`Context` should be surrounded by an `if` to avoid the case where `ty_params` is empty.

lib/src/top_level.rs
```diff
@@ -1629,32 +1658,53 @@ impl TopLevelItem {
...
-                                                line(),
-                                                ty_param.to_doc(false),
-                                                text(")"),
-                                            ])
-                                        } else {
-                                            ty_param.to_doc(false)
-                                        }),
-                                    ])
-                                })),
+                                // Below we want to make a list of assigned params, and we have to concat them together eventually
```
clarus, line 1666: OK top for the comments!

clarus:
I am surprised that the CI does not trigger for you, maybe this is because your branch has conflicts?

clarus:
So the current error in the CI requires to run:
```sh
cargo fmt
```

clarus:
Closing as the follow-up PR #100 is merged.

PR #96, opened 2023-06-19
Handle `const _  = ...` expressions

### LINKED ISSUE #95: Remove the Definition unit
Understand where the:
```coq
Definition _ : unit := run (tt).
```
come from in the translation and remove them/translate them correctly, depending on the issue.

dhilst:
fix #95


dhilst:
I ran this command to expand the ink macros and see the generated code: `cargo expand --manifest-path ./integration-tests/flipper/Cargo.toml > integration-tests/flipper/flipper_expanded.rs` and then inspected the file.

image

We can see these `const _ : () = ...` in the generated code, these are the `Definition _ : unit := ...` in `.v` files.

dhilst:
Fixing the CI

I noticed that not all `const _ : ()` expressions have the body `run (Pure tt)`, some of them have calls to ink functions. It seems to me that these are some side effects, but didn't look that into the ink code

image

CoqOfRust/ink/erc20.v
```diff
@@ -3578,29 +3448,6 @@ Module Impl_erc20_erc20_Erc20.
...
-    let* _ := ink.codegen.utils.identity_type.consume_type tt in
-    let* _ := ink.codegen.utils.identity_type.consume_type tt in
-    let* _ := ink.codegen.utils.identity_type.consume_type tt in
-    let* _ := ink.codegen.utils.identity_type.consume_type tt in
-    let* _ := ink.codegen.utils.identity_type.consume_type tt in
-    let* _ := ink.codegen.utils.identity_type.consume_type tt in
-    let* _ := ink.codegen.utils.identity_type.consume_type tt in
-    let* _ := ink.codegen.utils.identity_type.consume_type tt in
-    let* _ := ink.codegen.utils.identity_type.consume_type tt in
-    Pure tt).
```
dhilst, line 3600: @clarus do you that removing this `ink.codegen.utils.identity_type.consume_type` is a problem?

clarus: Ok yes, seems sound to remove it according to the comment! I do not know if it is possible to have `const` that are actually making side effects. The way we will design the `run` command is to be a partial function that also takes a proof certificate (`eq_refl`) saying that the parameter is pure so that we can return an `A` from an `M A`.

Issue #90, opened 2023-06-15
Env: translate the Environment trait


clarus:
In https://github.com/paritytech/ink/blob/2eba99ee445a4aee34d0919041388e7e727f7ed8/crates/env/src/types.rs there is the definition of the trait `Environment`:
```rust
pub trait Environment {
    /// The maximum number of supported event topics provided by the runtime.
    ///
    /// The value must match the maximum number of supported event topics of the used
    /// runtime.
    const MAX_EVENT_TOPICS: usize;

    /// The account id type.
    type AccountId: 'static
        + scale::Codec
        + CodecAsType
        + Clone
        + PartialEq
        + Eq
        + Ord
        + AsRef<[u8]>
        + AsMut<[u8]>;
        
...
```
It is translated for now as:
```coq
Module Environment.
  Class Trait (Self : Set) : Set := {
    MAX_EVENT_TOPICS : usize;
    AccountId : Set;

...
```
in https://github.com/formal-land/coq-of-rust/blob/main/CoqOfRust/ink/Env.v

Instead, we want to have types that are "synonym fields" (with `:=` instead of `:`), and all the types being given as parameters together with the type-class information:
```coq
Module Environment.
  Class Trait (Self : Set) {AccountId : Set}
      `{scale.Codec AccountId}
      `{CodecAsType AccountId}
      `{Clone AccountId}
      ...
      : Set := {
    MAX_EVENT_TOPICS : usize;
    AccountId := AccountId;

...
```

dhilst:
The `inv_env.v` file after the #99 MR, the parameters 

![image](https://github.com/formal-land/coq-of-rust/assets/392820/8e8ca947-84a9-4036-9749-b353b3fd79a7)

PR #99, opened 2023-06-26
Add trait parameters to the record types during the translation

dhilst:
Part of #90


coq_translation/examples/traits/traits_parms.v
```diff
@@ -0,0 +1,68 @@
...
+      Set
+      :=
+      {
+    SomeType := SomeType;
+    some_fn : unit -> (M unit);
+  }.
+  
+  Global Instance Method_SomeType `(Trait) : Notation.Dot "SomeType" := {
+    Notation.dot := SomeType;
+  }.
```
dhilst, line 40: I need to do this modification for it to typecheck

```suggestion
  Global Instance Method_SomeType `(Trait) (x : SomeType) : Notation.Dot "SomeType" := {
    Notation.dot := x;
  }.
```

otherwise, I have this error 

```
The term "?Trait.(SomeType)" has type "Set" while it is expected to have type "?T"
(unable to find a well-typed instantiation for "?T": cannot ensure that "Type" is a subtype of 
"Set").
```

dhilst: I'm not sure if this is the right thing to do, or if I messed up with something without notice

dhilst: Maybe the right thing is to not produce this instance, as `foo.SomeType` does not make sense in rust

clarus: Oh, OK. Indeed, we can avoid to generate these cases, unless if we see it required at some point. If that the case, we will make the notation `Notation.Dot` more generic to also allow to return a type.

lib/src/top_level.rs
```diff
@@ -1481,77 +1518,83 @@ impl TopLevelItem {
...
-                                    text("Global Instance"),
-                                    line(),
-                                    text(format!("Method_{name}")),
-                                    line(),
-                                    text("`(Trait)"),
-                                ]),
-                                line(),
+                        if matches!(item, TraitItem::Type(..)) {
+                            nil()
+                        } else {
```
dhilst, line 1523: I added this `if` to skip the `Notation.Dot` for a `SomeType`, the changed lines below were indented

coq_translation/examples/traits/traits_parms.v
```diff
@@ -0,0 +1,61 @@
...
+  Unset Primitive Projections.
+  Class Trait (Self : Set) : Set := {
+  }.
+  Global Set Primitive Projections.
+End Tar.
+
+Module SomeTrait.
+  Class Trait (Self : Set) : Set := {
+    SomeType : Set;
+    some_fn :  -> (M unit);
```
dhilst: Fixed in the next commit

lib/src/ty.rs
```diff
@@ -141,10 +141,14 @@ impl CoqType {
...
-                        [text(" ->"), line()],
-                    ),
+                    if args.is_empty() {
+                        text("unit")
+                    } else {
+                        intersperse(
+                            args.iter().map(|arg| arg.to_doc(true)),
+                            [text(" ->"), line()],
+                        )
+                    },
```
dhilst, line 151: Fixes https://github.com/formal-land/coq-of-rust/pull/99/commits/bacec7775b12bf4dee7cffc436ca1e86468612bf#r1243623216

lib/src/top_level.rs
```diff
@@ -426,7 +426,19 @@ fn compile_top_level_item(tcx: &TyCtxt, env: &mut Env, item: &Item) -> Vec<TopLe
...
+                                    .iter()
+                                    .filter_map(|generic_bound| match generic_bound {
+                                        GenericBound::Trait(ptraitref, _) => {
+                                            Some(compile_path(env, ptraitref.trait_ref.path))
+                                        }
+                                        GenericBound::LangItemTrait { .. } => None,
+                                        GenericBound::Outlives { .. } => None,
+                                    })
+                                    .collect();
+                                TraitItem::Type(generic_bounds)
```
dhilst, line 440: This is where I get the trait bounds, 
image

I compile each one as a path

lib/src/top_level.rs
```diff
@@ -1435,14 +1447,39 @@ impl TopLevelItem {
...
+                                            group([
+                                                text(" `{"),
+                                                x.to_doc(),
+                                                text(".Trait"),
+                                                text(" "),
+                                                text(item_name),
+                                                text("}"),
+                                                line(),
+                                            ])
+                                        })),
```
dhilst, line 1474: And here I use the trait bounds 
image

dhilst:
The CI is failing because of the translated erc.v has changes that were not yet committed, I'm looking into it

coq_translation/examples/traits/traits_parms.v
dhilst: I'm not generating this instance anymore

Issue #89, opened 2023-06-15
Env: translate the caller function


clarus:
Depends on https://github.com/formal-land/coq-of-rust/issues/90

In https://github.com/paritytech/ink/blob/2eba99ee445a4aee34d0919041388e7e727f7ed8/crates/env/src/api.rs there is the Rust code:
```rust
pub fn caller<E>() -> E::AccountId
where
    E: Environment,
{
    <EnvInstance as OnInstance>::on_instance(|instance| {
        TypedEnvBackend::caller::<E>(instance)
    })
}
```
that we (incorrectly) translate to:
```coq
Definition caller
    {E : Set}
    `{Environment.Trait E}
    (_ : unit)
    : M ImplE.AccountId :=
  OnInstance.on_instance (fun instance => TypedEnvBackend.caller instance).
```

We should:

- [ ] remove the `(_ : unit)` parameter
- [ ] translate `E::AccountId` correctly

InfiniteEchoes:
`(_ : unit)` happens when there's no arguments for a function. Does it mean that we should set it to nil when there're no arguments?

clarus:
> `(_ : unit)` happens when there're no arguments for a function. Does it mean that we should set it to nil when there're no arguments?

Yes, we should have an empty list of arguments when there are no arguments.

clarus:
Not a priority anymore.

PR #101, opened 2023-06-27
Caller Function

InfiniteEchoes:
This PR is for #89 .


clarus:
You can re-generate the files by running the commands that are the README, namely:
```sh
cargo run --bin coq-of-rust -- translate --path examples
python run_tests.py
```

The second command uses Python and does essentially the same as the first one, but also saves the error messages. So in general the `cargo` command is enough and faster.

clarus:
OK so I updated the snapshots, the function calls should also be updated to remove the spurious `tt` parameter!

InfiniteEchoes:
Due to frequent updates in the main branch, I messed it up when rebasing this PR from newest main branch. This PR is polluted and closed, and I'll replace with a new PR.

PR #106, opened 2023-07-03
Add Default, AsRef and AsMut traits

dhilst:


dhilst:
Closing this for now (as these traits already exists in `std/`

Issue #104, opened 2023-07-03
Compile the beginning of ink_env


clarus:
The beginning of https://github.com/formal-land/coq-of-rust/blob/main/CoqOfRust/ink/ink_env.v starts with:
```coq
(* Generated by coq-of-rust *)
Require Import CoqOfRust.CoqOfRust.

Module types.
  Module FromLittleEndian.
    Class Trait
        (Self : Set)
        {Bytes : Set}
        `{core.default.Default.Trait Bytes}
        `{core.convert.AsRef.Trait Bytes}
        `{core.convert.AsMut.Trait Bytes} :
        Set := {
      Bytes := Bytes;
      from_le_bytes : ImplSelf.Bytes -> (M Self);
    }.
```

Make it compile in Coq, probably plugging/adding the missing traits in the standard library.

InfiniteEchoes:
My observations: 

The module being imported is `CoqOfRust.CoqOfRust`. For the `std` part that I have designed, all of them are under `CoqOfRust.std`, and auxiliary definitions are defined in `CoqOfRust.lib.lib`. So every modules under `std` wouls start with a line of `Require Import CoqOfRust.lib.lib.`.

`CoqOfRust.CoqOfRust` should have overlapped/outdated content with `CoqOfRust.lib.lib`, in the auxiliary definitions(one issue here is whether we can unite the `CoqOfRust.v` and `lib.v` together). Merely import this file would not import modules like `core` and `std`. I wonder if we might need lines like `Require core` or equivalent.

dhilst:
Oh, so we have the Traits, I was about to add those, thanks! I'll look at how to integrate CoqOfRust, lib and std

dhilst:
The #108 files should add the stdlib dependencies that were missing to compile the start of ink_env but there is more stuff to do yet.

PR #108, opened 2023-07-03
Use std files

dhilst:
- Rename std to _std to avoid module shadowing warnings
- Rename imports in _std files `s/std/_std/` 
- Move monad to its own file to avoid circular dependency
- Blacklist one example not compiling
- Update the _std files with the content of `std` module in CoqOfRust (which was being used)
- Remove `std` module from CoqOfRust.v
- Require `Monad` file in some _std file where needed


CoqOfRust/ink/erc20.v
```diff
@@ -3549,6 +3549,80 @@ Module Impl_parity_scale_codec_codec_Decode_for_erc20_erc20___CallBuilder.
...
+  Definition decode_into
+      (__codec_input_edqy : mut_ref __CodecInputEdqy)
+      (dst_ : mut_ref (core.mem.maybe_uninit.MaybeUninit Self))
+      :
+        M
+          (core.result.Result
+            parity_scale_codec.decode_finished.DecodeFinished
+            parity_scale_codec.error.Error) :=
+    let* _ :=
+      let* α0 := core.mem.size_of tt in
```
dhilst, line 3561: I copied this change from the error in the CI, but if I ran `./generate_ink_example.sh` it does not change anything in `erc20.v` file, I don't understand why this change is showing up in the CI

dhilst:
parti of #104

PR #125, opened 2023-07-13
fixing the translation of flow_of_control/

bartlomiejkrolikowski:


CoqOfRust/CoqOfRust.v
```diff
@@ -409,6 +419,26 @@ Module core.
...
+
+    Parameter assert_failed :
+      forall `{State.Trait} {T U : Set} `{fmt.Debug.Trait T} `{fmt.Debug.Trait U},
+      AssertKind -> ref T -> ref U -> option.Option fmt.Arguments -> M Empty_set.
+
+(*     Definition assert_failed
+      {T U : Set} `{fmt.Debug.Trait T} `{fmt.Debug.Trait U}
+      (kind : AssertKind) (left : ref T) (right : ref U)
+      (args : option.Option fmt.Arguments) : Empty_set :=
+        assert_failed_inner kind left right args. *)
```
clarus, line 439: Without comment for the final version of the pull-request.

CoqOfRust/CoqOfRust.v
```diff
@@ -1330,3 +1372,180 @@ Definition format_argument : Set := core.fmt.ArgumentV1.
...
+  Admitted.
+(*   Impl_Iterator_for_Range.Method_next (A := Z). *)
+End Impl_Iterator_for_Range_Z.
+
+(* temporary argument specifications
+ * it would probably be better if they were defined in the _std files
+ *)
+
+Arguments std.option.Option.Some {_} _.
+Arguments std.option.Option.None {_}.
```
clarus, line 1551:
```suggestion
```
I added it next to the definition of `Option`.

CoqOfRust/_std/slice.v
```diff
@@ -51,7 +51,7 @@ where
 { /* private fields */ }
 *)
 Module ArrayChunksMut.
-  Parameter t : Set -> usize -> Set.
+  Record t (T : Set) (N : usize) : Set := { }.
```
clarus, line 54: I think we can keep the definition with `Parameter` as the fields are private. This also avoids having an empty record that makes a compilation warning for primitive projections. Or you have a reason to do this change for later?

bartlomiejkrolikowski: Ok.
This is not my change. It may be just a not updated file.

CoqOfRust/CoqOfRust.v
bartlomiejkrolikowski: done

CoqOfRust/CoqOfRust.v
bartlomiejkrolikowski: done

CoqOfRust/_std/slice.v
bartlomiejkrolikowski: done

PR #131, opened 2023-07-18
Caller Function(New branch)

InfiniteEchoes:
This PR is for #89 , as a new substitution for the last broken PR. This PR is planned to:

- [x] Eliminated redundant `(_ : unit)`
- [x] Eliminated redundant `tt` parameters after a function call
- [ ] Translate the return type correctly

When eliminating the redundant `tt`, I find that functions with generic type parameters might miss their type params.


lib/src/expression.rs
```diff
@@ -912,11 +912,13 @@ impl Expr {
...
                     func.to_doc(true),
-                    line(),
                     if args.is_empty() {
-                        text("tt")
+                        nil()
                     } else {
-                        intersperse(args.iter().map(|arg| arg.to_doc(true)), [line()])
+                        concat([
+                            line(),
+                            intersperse(args.iter().map(|arg| arg.to_doc(true)), [line()]),
```
clarus, line 920: I think it is better to do `concat` rather than `intersperse` once we remove the empty case with `unit`.

clarus:
Closing as changes included in https://github.com/formal-land/coq-of-rust/pull/129

PR #133, opened 2023-07-18
Reorder files

dhilst:
* Add `coq-of-rust-config.json` file
* A `reorder` section can be used to reorder definitions in various scopes
    
    ```json
    {
      "reorder": {
        "examples/functions/functions_order.rs": {
          "top_level": [
            "SomeType",
            "Impl_core_default_Default_for_functions_order_SomeType",
            "Impl_functions_order_SomeType",
            "inner_mod",
            "foo",
            "main"
          ],
          "top_level::Impl_core_default_Default_for_functions_order_SomeType": ["default"],
          "top_level::Impl_functions_order_SomeType": ["meth2", "meth1"],
          "top_level::inner_mod": ["tar", "bar"],
          "top_level::inner_mod::nested_mod": ["tack", "tick"]
        }
      }
    }
    ```

* Add a `file` a and `context` to the `Env` type. `file` is the file being compiled and `context` is the context. We use this to find the proper section the json file 
* Add a trait to get names from various AST objects
* Add a function to reorder a vector of definitions in place
* Add two method to push and pop contexts during the compilation, this is used to track the compilation scope
* Add `--configuration-file` option to specify the configuration file, the default is `coq-of-rust-config.json`
* Add `--generate-reorder` to print the reorder section of the configuration file, this is to avoid having to figure out the symbol names
  
image

*


dhilst:
There is a subtle bug visible in the output json 
image

Is it possible to see `top_level::nested_mod` in the output which is not a valid scope in the `functions_order.rs` the correct is `top_level::inner_mod::nested_mod` which also shows up in the output. The `reorder_definitions_inplace` is called inside `compile_top_level_item` recursively, surrounded by `env.push_context(name)` and `env.pop_context()`, so in some way, the call order is not exactly what I was expecting, maybe because of the transformations calling compile again!? (it's a guess)

dhilst:
> There is a subtle bug visible in the output json image
> 
> Is it possible to see `top_level::nested_mod` in the output which is not a valid scope in the `functions_order.rs` the correct is `top_level::inner_mod::nested_mod` which also shows up in the output. The `reorder_definitions_inplace` is called inside `compile_top_level_item` recursively, surrounded by `env.push_context(name)` and `env.pop_context()`, so in some way, the call order is not exactly what I was expecting, maybe because of the transformations calling compile again!? (it's a guess)

Okay, I got what is wrong.

I assumed that the compilation went recursively respecting the structure of the HIR, and this is true, *but* we also iterate over the nodes using `tcx.hir().items().flat_map(...)` which iterate over HIR nodes linearly, this is why I see the buggy scopes in the output of configuration file.

When we call `compile_top_level` for the first time, it calls `compile_top_level_item` inside a `flat_map`, when this call of `compile_top_level_item` see a module, it calls itself, so the nested items are compiled first. When we have a nested AST, `compile_top_level_item` will be called recursively first, then `flat_map` iterator will reach the nested AST nodes later, and
`compile_top_level_item` is called a second time with the same item in this second case, with `context` in a buggy state because the call stack does not follow the recursive structure of the AST

clarus (approved):

PR #134, opened 2023-07-20
still fixing the translation of flow_of_control/

bartlomiejkrolikowski:


CoqOfRust/CoqOfRust.v
```diff
@@ -1408,6 +1423,90 @@ Module Impl_Slice.
...
+  Definition Item := A.
+
+  Parameter next :
+    forall `{State.Trait}, mut_ref Self -> M (core.option.Option A).
+
+  Global Instance Method_next `{State.Trait} : Notation.Dot "next" := {
+    Notation.dot := next;
+  }.
+  End Impl_Iterator_for_Slice_Iter.
+End Impl_Iterator_for_Slice_Iter.
```
dhilst, line 1443: Why not on _std files?

dhilst (approved):

CoqOfRust/CoqOfRust.v
bartlomiejkrolikowski: It may be good to move this definition there.
I wasn't sure how to make changes in those files and whether they should be modified manually,
so, just in case, put all the new definitions here.

PR #147, opened 2023-07-26
136 support records with type parameters

### LINKED ISSUE #136: Support records with type parameters
Successfully translate the type definition:
```rust
struct PhantomTuple<A, B>(A, PhantomData<B>);
```
in https://github.com/formal-land/coq-of-rust/blob/main/examples/generics/generics_phantom_type.rs For now it gets translated to:
```coq
Record t : Set := { _ : A; _ : core.marker.PhantomData B;}.
```
to the type parameters `A` and `B` are missing.

bartlomiejkrolikowski:
closes #136


lib/src/render.rs
```diff
@@ -157,3 +157,64 @@ where
...
+pub(crate) fn add_section_if_necessary<'a, U>(name: U, ty_params: &Vec<U>, doc: Doc<'a>) -> Doc<'a>
+where
+    U: Into<std::borrow::Cow<'a, str>> + std::marker::Copy,
+{
+    if ty_params.is_empty() {
+        doc
+    } else {
+        section(name, ty_params, doc)
+    }
+}
```
bartlomiejkrolikowski, line 220: I needed to make the same change a few times, so I defined those function to make it less complicated.
The unwanted result is spaces in empty lines, but I couldn't fix that.

clarus: Yes, spaces in empty lines are always there. We will eventually add a last-step filter pass to remove them.

CoqOfRust/blacklist.txt
```diff
@@ -62,17 +62,13 @@ examples/functions/functions.v
...
 examples/generics/generics_bounds.v
-examples/generics/generics_functions.v
 examples/generics/generics_implementation.v
 examples/generics/generics_multiple_bounds.v
 examples/generics/generics_new_type_idiom.v
 examples/generics/generics_phantom_type_test_case_unit_clarification.v
 examples/generics/generics_phantom_type.v
-examples/generics/generics_traits.v
 examples/generics/generics_where_clauses.v
-examples/generics/generics.v
```
clarus, line 75: :tada:

lib/src/top_level.rs
```diff
@@ -385,8 +388,26 @@ fn compile_top_level_item(tcx: &TyCtxt, env: &mut Env, item: &Item) -> Vec<TopLe
...
+                .filter_map(|param| match param.kind {
+                    rustc_hir::GenericParamKind::Type { .. } => {
+                        Some(param.name.ident().to_string())
+                    }
+                    _ => {
+                        env.tcx
+                            .sess
+                            .struct_span_warn(param.span, "Only type parameters are supported.")
+                            .note("It should be supported in future versions.")
+                            .emit();
```
clarus, line 405: We should always ignore lifetime parameters.

lib/src/render.rs
```diff
@@ -157,3 +157,64 @@ where
...
+            } else {
+                let types: Vec<Doc<'a>> = ty_context.iter().map(|&ty| text(ty)).collect();
+                group([
+                    nest([
+                        text("Context"),
+                        line(),
+                        text("["),
+                        intersperse(types, [line()]),
+                        line(),
+                        text(": Set]."),
```
clarus, line 182:
```suggestion
                        text("{"),
                        intersperse(types, [line()]),
                        line(),
                        text(": Set}."),
```
using `{ ... }` for implicit for uniformity with the rest, if it works.

bartlomiejkrolikowski: I checked that neither `{ ... }` nor `( ... )` work for `generics/generics.v`. At least not instantly.

bartlomiejkrolikowski: So I did that because I thought it will be faster that way.

bartlomiejkrolikowski: but now I see `{ ... }` work when I prepend the items with `@` in the definitions of aliases,
so I'll modify it this way

lib/src/top_level.rs
bartlomiejkrolikowski: I removed the warning.
But can't there also be constant parameter? Shouldn't we have a warning for that?

PR #149, opened 2023-07-27
137 add implicit parameter when calling traits

### LINKED ISSUE #137: Add implicit parameter when calling traits
The trait `core.cmp.PartialEq.Trait` has one necessary type parameter and one implicit. Currently, when this trait is implemented we get:
```coq
Global Instance I : core.cmp.PartialEq.Trait Self := {
  core.cmp.PartialEq.eq `{H : State.Trait} := eq;
}.
```
but we should have with the implicit type parameter at `None`:
```coq
Global Instance I : core.cmp.PartialEq.Trait Self None := {
  core.cmp.PartialEq.eq `{H : State.Trait} := eq;
}.
```

bartlomiejkrolikowski:
closes #137

### DIFF
--- a/CoqOfRust/core/cmp.v
+++ b/CoqOfRust/core/cmp.v
@@ -32,7 +32,7 @@ Traits
 [x] PartialOrd
 *)
 Module PartialEq.
-  Class Trait (Self : Set) (Rhs : option Set) : Set := {
+  Class Trait (Self : Set) {Rhs : option Set} : Set := {
     Rhs := defaultType Rhs Self;
 
     eq `{State.Trait} : ref Self -> ref Rhs -> M bool;
--- a/lib/src/top_level.rs
+++ b/lib/src/top_level.rs
@@ -6,8 +6,8 @@ use crate::render::*;
 use crate::ty::*;
 use rustc_ast::ast::{AttrArgs, AttrKind};
 use rustc_hir::{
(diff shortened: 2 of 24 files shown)


CoqOfRust/core/cmp.v
```diff
@@ -32,7 +32,7 @@ Traits
 [x] PartialOrd
 *)
 Module PartialEq.
-  Class Trait (Self : Set) (Rhs : option Set) : Set := {
+  Class Trait (Self : Set) {Rhs : option Set} : Set := {
```
bartlomiejkrolikowski, line 35: To apply the optional argument by assignment to the proper name.
(i.e. `... (Rhs := None)`)

lib/src/top_level.rs
```diff
@@ -1756,11 +1786,18 @@ impl TopLevelItem {
...
+                                                    // @TODO: implement translation of type parameters with default
+                                                    Some(_default) => concat([
+                                                        text("(* TODO *)"),
+                                                        line(),
+                                                        text(ty),
+                                                        line(),
+                                                    ]),
+                                                    None => concat([text(ty), line()]),
+                                                }
+                                            })),
```
bartlomiejkrolikowski, line 1800: I made this to make sure we don't translate traits with parameters with default. That is because we don't translate the defaults yet.

Issue #138, opened 2023-07-21
In axiom mode, only translate public modules


clarus:
Only public modules/functions/data types should be translated in axiom mode, as the rest will not be used anyways (no private of `pub(crate)`).

PR #151, opened 2023-07-28
Only Translate Public Modules in Axiomatized Mode

InfiniteEchoes:
This PR is for #138 :
- When in axiomatize mode, only public item will be translated. 

Note: The most representative example for this issue is `modules/visibility.rs`. For now, I only check if the corresponded items have public access. I see that some other places might use something like relative visibility, in which you check if an item is visible with regard to some other item. I wonder if this will affect the result.


clarus:
Merged in the follow-up pull-request: https://github.com/formal-land/coq-of-rust/pull/152

PR #164, opened 2023-08-01
34 translate try from and try into.rs

### LINKED ISSUE #34: Translate try_from_and_try_into.rs
Translate the file https://github.com/formal-land/coq-of-rust/blob/main/examples-from-rust-book/conversion/try_from_and_try_into.rs

bartlomiejkrolikowski:
closes #34


CoqOfRust/CoqOfRust.v
```diff
@@ -53,30 +53,8 @@ Notation "e (||)" :=
...
-  Arguments double_colon {Kind} type name {T DoubleColon}.
-
-  (* A class to represent types in a trait. *)
-  Class DoubleColonType {Kind : Type} (type : Kind) (name : string) : Type := {
-    double_colon_type : Set;
-  }.
-  Arguments double_colon_type {Kind} type name {DoubleColonType}.
-End Notation.
+Require CoqOfRust.lib.lib.
+Module Notation := CoqOfRust.lib.lib.Notation.
```
bartlomiejkrolikowski, line 57: I defined some instances in the library files using the `Notation` module from `lib`, so I needed removed this one to help Coq guess the correct type associated with `::[""]` notations.

CoqOfRust/CoqOfRust.v
```diff
@@ -1369,6 +1414,28 @@ Definition format_arguments : Set := core.fmt.Arguments.
 
 Definition Slice := lib.slice.
 
+(* this is a specialized instance to make try_from_and_into.v work *)
+Global Instance Formatter_debug_tuple_field1_finish_for_i32 `{State.Trait} :
+  Notation.DoubleColon core.fmt.Formatter "debug_tuple_field1_finish" :=
+    core.fmt.ImplFormatter.Formatter_debug_tuple_field1_finish (T := i32).
```
bartlomiejkrolikowski, line 1420: Unfortunately Coq cannot guess proper type parameters of an instance and the generic instance of `Notation.DoubleColon core.fmt.Formatter "debug_tuple_field1_finish"` didn't work, so I had to create this specialized version (and it may be necessary to create specialized instances for every type `T` can be, or at least for every file that utilizes this function).

CoqOfRust/CoqOfRust.v
```diff
@@ -1369,6 +1414,28 @@ Definition format_arguments : Set := core.fmt.Arguments.
...
+
+    Parameter fmt :
+      forall `{State.Trait}, ref (core.result.Result T E) ->
+        mut_ref core.fmt.Formatter -> M core.fmt.Result.
+
+    Global Instance I : core.fmt.Debug.Trait (core.result.Result T E) := {
+      fmt `{State.Trait} := fmt;
+    }.
+  End Impl_Debug_for_Result.
+End Impl_Debug_for_Result.
```
bartlomiejkrolikowski, line 1437: `core.fmt` is defined in this file, so I weren't able to define this in `core.result"

CoqOfRust/core/result.v
```diff
@@ -1,48 +1,17 @@
-Require Import CoqOfRust.lib.lib.
+(* To avoid circular dependency
+ * the translation of core::result.rs is split into two files
+ *)
```
bartlomiejkrolikowski, line 3: To be able to define implementations of traits from `result.rs` in `result.v` I split it into `result_types.v` and `result_impl.v`.
Here I only aliased the names so that the change were invisible in other files.

lib/src/expression.rs
```diff
@@ -136,6 +136,8 @@ pub(crate) enum Stmt {
     Let {
         is_monadic: bool,
         pattern: Box<Pattern>,
+        /// optional type annotation (should be None for monadic lets)
+        ty: Option<Box<CoqType>>,
```
bartlomiejkrolikowski, line 140: When I defined the instance of `Debug.Trait` for `Result` some files stopped working, probably because there two different instances of this trait and Coq was choosing the wrong one. This was especially happening when `Option.None` was assingned to a variable in a `let` expression - Coq was incorrectly guessing the type parameter.
That's why I added the type annotations.

lib/src/path.rs
```diff
@@ -117,18 +117,22 @@ pub(crate) fn compile_qpath(env: &Env, qpath: &QPath) -> Path {
...
+                        let mut path = compile_path(env, path);
+                        path.prefix_last_by_impl();
+                        Some(path)
+                    }
+                },
+                _ => Some(Path::local("ComplexTypePath".to_string())),
+            };
+            let raw_segments = match ty {
+                Some(path) => vec![path.to_string(), segment.ident.name.to_string()],
+                None => vec![segment.ident.name.to_string()],
```
bartlomiejkrolikowski, line 132: This change is because the name of `SelfTyAlias` is `"Self"` and in this case names refer to objects already visible in the scope of the module.

coq_translation/default/examples/custom_types/structures.v
```diff
@@ -122,7 +122,7 @@ Definition main `{H : State.Trait} : M (H := H) unit :=
 " ]) (addr_of [ α0 ]) in
       std.io.stdio._print α1 in
     Pure tt in
-  let point :=
+  let point : structures.Point :=
     {| structures.Point.x := 10 (* 10.3 *); structures.Point.y := 0 (* 0.4 *);
     |} in
   let* _ :=
```
dhilst, line 128: I like that, is good to know the types, I don't know the impact on the file size but we'll see

CoqOfRust/core/result.v
```diff
@@ -1,48 +1,17 @@
...
-*)
-Module Iter.
-  Parameter t : Set -> Set.
-End Iter.
-Definition Iter := Iter.t.
+Module IntoIter := IntoIter.
+Module Iter := Iter.
+Module IterMut := IterMut.
+Module Result := Result.
+Module Impl_PartialEq_for_Result := Impl_PartialEq_for_Result.
```
dhilst, line 12: Isn't this redundant?

CoqOfRust/CoqOfRust.v
dhilst: Update the comment with the reason for this instance, say that there was a type inference error or something like that

CoqOfRust/core/convert.v
```diff
@@ -93,10 +93,15 @@ pub trait TryFrom<T>: Sized {
...
 Module TryFrom.
-  Class Trait (Self T Error : Set) : Set := { 
+  Class Trait (Self : Set) {T Error : Set} : Set := {
     Error := Error;
 
-    try_from : T -> Result Self Error;
+    try_from `{State.Trait} : T -> M (Result Self Error);
+  }.
+
+  Local Instance AssociatedFunction_try_from {Self : Set} `{Trait Self} :
```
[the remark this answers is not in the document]
bartlomiejkrolikowski: Sorry. I changed it to test something and forgot to change back.

CoqOfRust/CoqOfRust.v
bartlomiejkrolikowski: OK
I did it.

CoqOfRust/core/result.v
bartlomiejkrolikowski: I think it is necessary when we refer to the functions defined in those modules with fully qualified names.
I think otherwise we would need to write `core.result.result_type.Result.Ok` instead of `core.result.Result.Ok`.

CoqOfRust/core/convert.v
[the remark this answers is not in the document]
bartlomiejkrolikowski: I repaired it.

clarus (approved):

Issue #162, opened 2023-08-01
Handle polymorphic aliases


clarus:
In the file https://github.com/formal-land/coq-of-rust/blob/main/CoqOfRust/ink/ink_primitives.v , the translation:
```coq
Definition MessageResult : Set := core.result.Result T ink_primitives.LangError.
```
has a type parameter `T` that is not introduced, like:
```coq
Definition MessageResult (T : Set) : Set := core.result.Result T ink_primitives.LangError.
```

The goal of this task is to correctly handle the polymorphic parameters.

PR #169, opened 2023-08-03
Add generics for type aliases

InfiniteEchoes:
This PR is for #162 :
- Adds generic type parameters for type aliases
- Removes from blacklist two examples that compiles after the change.


lib/src/top_level.rs
```diff
@@ -1629,11 +1648,29 @@ impl TopLevelItem {
...
+                    } else {
+                        concat([
+                            line(),
+                            text("("),
+                            concat(ty_params.iter().map(|ty| concat([text(ty), line()]))),
+                            text(":"),
+                            line(),
+                            text("Set"),
+                            line(),
+                            text(")"),
```
clarus, line 1671:
```suggestion
                            text("Set)"),
```

clarus: To avoid having a space between `Set` and `)`.

clarus:
There are conflicts on the generated files.

PR #177, opened 2023-08-08
Rendering functions for traits

bartlomiejkrolikowski:


lib/src/render.rs
```diff
@@ -310,3 +368,174 @@ where
...
     ])
 }
+
+/// creates a body of a typeclass with the given items
+pub(crate) fn new_typeclass_body<'a, I>(items: I) -> Doc<'a>
+where
+    I: IntoIterator,
+    I::Item: pretty::Pretty<'a, pretty::RcAllocator, ()>,
+{
+    intersperse(items, [nil()])
```
clarus, line 378: I think it could be:
```suggestion
    concat(items)
```

PR #182, opened 2023-08-11
Compile ink_env.v

bartlomiejkrolikowski:
The continuation of PR #179.

PR #189, opened 2023-08-18
Compile more of ink env 2

bartlomiejkrolikowski:
The continuation of PR #182


lib/src/coq.rs
bartlomiejkrolikowski, line 1: Here I defined new types to control how certain Coq constructions are rendered.
I am testing them on the `Trait` case of `TopLevelItem::to_doc`.

CoqOfRust/_std/any.v
```diff
@@ -1,49 +1,10 @@
 Require Import CoqOfRust.lib.lib.
```
clarus, line 1: I do not think this file is required anymore (as calls to `std.any` should now be written `core.any` in the generated Coq code).

lib/src/coq.rs
clarus: OK top for the file to group these definitions.

scale-info
```diff
@@ -0,0 +1 @@
+../scale-info
```
clarus, line 1: I do not know if this link is required.

bartlomiejkrolikowski: Yes. I commited it by mistake.

CoqOfRust/_std/any.v
bartlomiejkrolikowski: Ok.
I removed it.

clarus (approved):

PR #191, opened 2023-08-23
Fix State.Trait in DoubleColonType bug

bartlomiejkrolikowski:
Fixes the problem with `State.Trait` parameters appearing in instances of `Notation.DoubleColonType`.


dhilst:
It makes sense to me

dhilst (approved):

PR #192, opened 2023-08-28
Compile more of ink_env.v 4

bartlomiejkrolikowski:
The continuation of PR #190


bartlomiejkrolikowski:
This PR applies changes to many files, but most of them are in one of the following categories:
- change `` `{H : State.Trait}`` to `` `{H' : State.Trait}``, because `H` was also used as a parameter at one place in Rust code
- stop generating parameters for associated types in `Trait` typeclasses
- add constraints of associated items to the fields of `Trait` typeclasses
- accordingly change the way associated types are obtaining an actual value in instance definitions
- add every instance `I` to the hint database
- use `eauto` to fill the constraint fields

clarus (approved):

PR #194, opened 2023-08-29
Fix From and Into examples

dhilst:


CoqOfRust/alloc/string.v
```diff
@@ -59,16 +59,16 @@ End ToString.
...
-  Global Instance FromStr : From.Trait str String := {
-    from := from;
+  Global Instance FromStr : From.Trait str (T := str) := {
+    from `{State.Trait} := from;
   }.
 
  Global Instance Method_from : Notation.DoubleColon String "from" := {
-    Notation.double_colon := from;
+    Notation.double_colon `{State.Trait} := from;
   }.
```
dhilst, line 72: I had to do this changes, it seems the interface that the translated code is expecting, but I'm not sure if the previous definition was outdated or if I'm missing something

clarus (approved):

PR #199, opened 2023-09-01
Compile more of ink_env.v 5

bartlomiejkrolikowski:
The continuation of #192.
Applies temporary changes that let ink_env.v be compiled.


lib/src/configuration.rs
```diff
@@ -14,14 +14,19 @@ pub(crate) struct DefinitionMove {
...
-            None => match &self.after {
-                Some(x) => x.clone(),
+            Some(x) => Some(x.clone()),
+            None => self.after.clone(),
+            /*original:
+            match &self.after {
+                Some(x) => Some(x.clone()),
                 None => panic!("Expecting before or after to be present in DefinitionMove"),
             },
+            */
```
bartlomiejkrolikowski, line 29: I changed this so that now items can be removed by not giving labels "before" and "after" in the JSON file.

clarus, line 29:
```suggestion
```

PR #225, opened 2023-09-29
Compile more of ink_env.v 13 - experimental

bartlomiejkrolikowski:
The continuation of #222.
Trying to use tactics to define instances of traits of associated types.


bartlomiejkrolikowski:
I tried to use tactics to automatically extract instances (with instances of related superclasses) related to associated types from other instances.
Unfortunatelly, in the current state my computer runs out of memory during compilation and I currently run out of ideas.

PR #381, opened 2023-10-06
Add definitions for structs' default parameters

InfiniteEchoes:


lib/src/top_level.rs
```diff
@@ -2534,6 +2534,28 @@ impl TypeStructStruct {
         // making fields immutable, just in case
         let fields = fields;
 
+        // List of parameters with default values
+        let params_with_default = ty_params
+            .iter()
+            .filter_map(|(ty, default)| {
+                if default.is_some() {
```
clarus, line 2541: With a `match` to avoid the `.unwrap()`.

clarus (approved):

PR #391, opened 2023-10-24
Draft: use Coq 8.18 in the CI

clarus:


clarus:
Closing as this way to install Coq (compiling from sources) is too slow (more than 20 minutes to compile Coq itself).

Issue #394, opened 2023-10-25
Remove core.marker.Sized.Trait


clarus:
The trait `core.marker.Sized.Trait` is used a lot but is only a marker (empty trait). Remove it automatically from the translation. This trait alone can complexify the inference of instances. Maybe this rule can easily be extended to all empty traits.

PR #395, opened 2023-10-27
Remove core.marker.Sized.Trait

InfiniteEchoes:
This PR is for #394 .
- Removed occurences of `core.marker.Sized.Trait`


lib/src/top_level.rs
```diff
@@ -1586,17 +1586,39 @@ impl ImplItem {
...
-        )
+  // gy@NOTE: This function might be able to generalize to more specific empty traits in general
+  fn is_sized_trait(path: &Path) -> bool {
+      let segments = path.segments.to_owned();
+      let sized_trait = vec!["core".to_string(), "marker".to_string(), "Sized".to_string(), "Trait".to_string()];
+      let matching = sized_trait.iter().zip(&segments).filter(|&(a, b)| a == b).count();
+      matching == sized_trait.len()
+  }
+
+  fn is_not_empty_trait(expr: &Box<coq::Expression>) -> bool {
```
clarus: The type `coq::Expression` is for printing only, so the analysis should be done before (on `TopLevelItem` for example, if it fits).

lib/src/top_level.rs
```diff
@@ -1586,17 +1586,39 @@ impl ImplItem {
...
-                    .collect(),
-            },
-            coq::ArgSpecKind::Implicit,
-        )
+  // gy@NOTE: This function might be able to generalize to more specific empty traits in general
+  fn is_sized_trait(path: &Path) -> bool {
+      let segments = path.segments.to_owned();
+      let sized_trait = vec!["core".to_string(), "marker".to_string(), "Sized".to_string(), "Trait".to_string()];
+      let matching = sized_trait.iter().zip(&segments).filter(|&(a, b)| a == b).count();
+      matching == sized_trait.len()
```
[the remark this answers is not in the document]
InfiniteEchoes: I believe `PartialEq` trait is not implemented for Vec<String>, as the compiler reported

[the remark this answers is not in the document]
clarus: It should work. The following code does type-check:
```rust
fn foo(v1: Vec<String>, v2: Vec<String>) -> bool {
    v1 == v2
}
```

lib/src/top_level.rs
```diff
@@ -301,6 +301,34 @@ fn is_top_level_item_public(tcx: &TyCtxt, env: &Env, item: &Item) -> bool {
...
+    ];
+    let matching = sized_trait
+        .iter()
+        .zip(segments)
+        .filter(|&(a, b)| a == b)
+        .count();
+    matching == sized_trait.len()
+}
+
+// gy@TODO: change the WherePredicate type into a general type T with a customizable function
```
clarus, line 319:
```suggestion
```
TODOs might never be done

lib/src/top_level.rs
```diff
@@ -357,6 +385,7 @@ fn compile_top_level_item(tcx: &TyCtxt, env: &mut Env, item: &Item) -> Vec<TopLe
             }
             let is_dead_code = check_dead_code_lint_in_attributes(tcx, item);
             let fn_sig_and_body = get_hir_fn_sig_and_body(tcx, fn_sig, body_id);
+            // gy@TODO: maybe move the filter to here
```
clarus, line 388:
```suggestion
```

clarus (approved):

PR #410, opened 2023-11-20
Add tool to generate contracts ready to be translated to Coq

clarus:


clarus:
The generated Rust code is valid, with a sandboxing of all the primitives to type-check without requiring the Ink! libraries. However, the generated Coq code does not compile yet. We will need to work on that.

PR #436, opened 2023-12-28
Fixes on `concurrent_tests.v`

InfiniteEchoes:
- Changed `Error` struct in `literals.v` to be under `error` module
- Changed `Range` struct in `core/ops.v` and `std/ops.v` to be under `ranged` module
- Added respective substitutions in `path.rs`

### DIFF
--- a/lib/src/path.rs
+++ b/lib/src/path.rs
@@ -170,6 +170,12 @@ pub(crate) fn to_valid_coq_name(str: &str) -> String {
         return format!("{}_", str);
     }
 
+    let reserved_struct_names = ["end"];
+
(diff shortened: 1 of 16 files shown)


lib/src/path.rs
```diff
@@ -170,6 +170,12 @@ pub(crate) fn to_valid_coq_name(str: &str) -> String {
         return format!("{}_", str);
     }
 
+    let reserved_struct_names = ["end"];
+
+    if reserved_struct_names.contains(&str) {
```
clarus, line 175: `"end"` can be added to the `reserved_names` list above.

PR #438, opened 2024-01-02
Escape special characters in strings

InfiniteEchoes:


CoqOfRust/examples/default/examples/cargo/concurrent_tests.v
```diff
@@ -22,8 +22,7 @@ Definition foo {A : Set} (o : core.option.Option.t A) : M unit :=
             let γ0 := γ.["Some.0"] in
             let* _a := M.copy γ0 in
             let* _ : M.Val unit :=
-              let* α0 : ref str.t := M.read (mk_str "some
-") in
+              let* α0 : ref str.t := M.read (mk_str "some\n") in
```
clarus, line 25: I do not think that the `\n` character is correctly interpreted by Coq as it has different rules than more strings in other languages. The syntax for strings in Coq is explained at the end of this file: https://coq.inria.fr/doc/master/stdlib/Coq.Strings.String.html

PR #450, opened 2024-01-22
Fix panics when getting THIR

clarus:
It is unclear to me what is the reason for some panic that we have when getting the THIR of some expressions, but at least this pull request wraps these errors properly.


clarus:
Fixes https://github.com/formal-land/coq-of-rust/issues/453

jmikedupont2:
Thanks for the quick turnaround!

PR #58, opened 2023-05-05
Add monadic translation

clarus:
Still draft.

PR #468, opened 2024-02-09
New monadic notation

olympichek:
Implement a representation of Rust programs without using explicit names for intermediate computation results. The notation is based on `M.run` markers (defined as a parameter with type `∀ {A : Set}, M A → A`) and the `M.monadic` tactic. `M.monadic` tactic replaces calls of `M.run` with calls to monadic bind (`let_`) operations.

`M.monadic` tactic is based on the `block` tactic from an older implementation of this idea (#58) with some adjustments. `e (| e1 , .. , en |)` notation for `M.run` is also taken from there.

`MonadicNotationExample.v` contains an example of a Rust program (simplified version of `rust_book/expressions/blocks.rs`) and its representation in Coq using old and new notations. It also serves as a test case for the `M.monadic` tactic. The code produced by `M.monadic` tactic is not exactly the same as the current translation result but is supposed to be equivalent up to monad laws (associativity of bind, identity of pure etc).

If this notation would be approved the next step would be to make `coq-of-rust` generate translation results using this notation. Then the `MonadicNotationExample.v` will no longer be relevant and can be removed.

### DIFF
--- a/CoqOfRust/M.v
+++ b/CoqOfRust/M.v
@@ -140,6 +140,13 @@ Module Option.
     end.
 End Option.
 
+(** This parameter is used as a marker to allow a monadic notation
+    without naming all intermediate results. Computation represented using
+    this markers can be translated to a regular monadic computation using
+    [M.monadic] tactic. Additionally, this parameter is used for the
+    definitions of "const".*)
+Parameter run : forall {A : Set}, M A -> A.
+
 Module Notations.
   Notation "'let-' a := b 'in' c" :=
     (LowM.let_ b (fun a => c))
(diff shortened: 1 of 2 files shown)


CoqOfRust/M.v
```diff
@@ -156,9 +163,43 @@ Module Notations.
...
 End Notations.
 Import Notations.
 
+(** A tactic that replaces all [M.run] markers with a bind operation.
+    This allows to represent Rust programs without introducing
+    explicit names for all intermediate computation results. *)
+Ltac monadic e :=
+  match e with
+  | context ctxt [let v : _ := ?x in @?f v] =>
+    refine (M.let_ _ _);
```
clarus, line 183: :+1:

clarus (approved):

PR #469, opened 2024-02-21
wip: Implement translation using new monadic notation

olympichek:
### Implement translation of the Rust programs using new monadic notation (#468)

Additionally, this pull request introduces various improvements to the `M.monadic` tactic as well as a new `M.monadic_match_operator` tactic that allows to use the new monadic notation for match expressions.

This pull request implements the following updates to the Rust translator:
* `mt_expression()` function that was used to perform monadic transformation of expressions is removed as it is no longer needed with the representation using new notation
* Generated Coq code for bodies of functions is wrapped in a call to the `M.monadic` tactic
* `M.monadic` tactic is also called on bodies of the `const` items
* Function calls that are marked as effectful use new notation: `e (| e1 , .. , en |)`
* Match expressions are translated to the call of the `M.monadic_match_operator` tactic instead of the `match_operator` function
* Types of the bodies of lambdas are no longer wrapped in a monad
* Lambdas that represent branches of match expressions have type annotations in their arguments

As well as the following updates on the side of Coq:
* `M.monadic` tactic now handles not only continuations of the `let .. in ..` expressions but also their contexts. Consider the following example: `f (let v := x in g v)`. I refer to `g` as a continuation and to `f` as a context
* `M.monadic` now generates a fresh variable when generating a lambda for a bind operation that is a result of translation of the `let .. in ..` expression. This fixes a problem related to variable shadowing in some examples
* `M.monadic` now wraps an expression in a call to `M.pure` when it is needed. This fixes a problem with examples where a plain expression (without any monadic operations and calls to `M.run`) is used at the end of the `let .. in ..` sequence.
* `lazymatch` is used within the `M.monadic` tactic. This does not change it's semantics but results with more relevant error messages when it does not (yet) work
* `M.monadic` now checks if the context is empty before inserting a call to the bind (`let_`) operation. Now it does not insert redundant calls to bind with a lambda and a call to `M.pure`. This results in shorter and cleaner terms
* New tactics: `M.monadic_match_branches` and `M.monadic_match_operator`. They allow representation of the match expressions using the new notation

Additionally:
* Example of the old representation was removed form `MonadicNotationExample.v`
* Example with the new representation of match expressions is added to `MonadicNotationExample.v`
* Snapshots of the translation results of examples are updated


clarus:
OK top, I do not have much to add!

clarus:
The current issue in the "hello world" example seems indeed doable with additional cases in the tactic.

PR #486, opened 2024-03-26
`expression::Literal` to_coq

InfiniteEchoes:
This PR:
- Added a `Paren` type in `coq::Expression` enum, for the purpose of decoupling
- Rewrote `string_to_doc` into `string_to_coq`. Future plan is to move this function out of `render.rs` to somewhere else
- Decoupled `Literal::to_doc` and rewrote into its `to_coq` version. Since the result would be later used in `expression::Expr`, the result has type `coq::Expression` rather than `coq::TopLevel`.

### DIFF
--- a/lib/src/coq.rs
+++ b/lib/src/coq.rs
@@ -168,6 +168,11 @@ pub(crate) enum Expression<'a> {
         /// a flag, set if implicit arguments are deactivated with '@'
         no_implicit: bool,
     },
+    Paren {
+        // An extra parameter for more recursive managements?
+        // with_paren: bool,
+        expr: Rc<Expression<'a>>,
+    },
     /// a wildcard: '_'
     Wild,
 }
@@ -593,6 +598,7 @@ impl<'a> Expression<'a> {
                 concat([optional_insert(!*no_implicit, text("@")), ident.to_doc()])
             }
             Self::Wild => text("_"),
+            Self::Paren { expr } => paren(true, expr.to_doc(with_paren)),
         }
     }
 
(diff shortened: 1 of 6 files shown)


lib/src/render.rs
```diff
@@ -89,46 +90,45 @@ fn cut_string_in_pieces_for_coq(input: &str) -> Vec<StringPiece> {
...
+                    format!("\"{}\"", str::replace(s, "\"", "\"\"")).as_str(),
+                );
+                if rest.is_empty() {
+                    head
+                } else {
+                    // gy@NOTE: both `apply` and `apply_many` would inherently add unecessary parens
+                    // since they're the only ways to represent at `coq::Expression` level, `apply_many`
+                    // is the way with least redundant parenthesis
+                    head.apply_many(&[
+                        coq::Expression::just_name("++"),
```
clarus, line 111: I think there should be a special combinator in `coq::Expression` for binary operators such as `++`, to avoid representing them with `apply_many`. That may simplify the parenthesis handling.

lib/src/coq.rs
```diff
@@ -168,6 +168,9 @@ pub(crate) enum Expression<'a> {
         /// a flag, set if implicit arguments are deactivated with '@'
         no_implicit: bool,
     },
+    Paren {
```
clarus, line 171:
```suggestion
    // TODO: remove to have parenthesis automatically detected
    Paren {
```
Adding a TODO comment

lib/src/render.rs
InfiniteEchoes: Do we treat it as a string concatenation or we set up a complete new combinator for binops?

clarus: We can setup a complete new combinator, with the parameters `lhs`, `rhs` for the two expressions, and `op`: `String` for the operator symbol.

clarus (approved):

Issue #270, opened 2023-10-04
Translate the example match_destructuring_arrays_slices.v


clarus:
Translate the file https://github.com/formal-land/coq-of-rust/blob/main/examples/flow_of_control/match_destructuring_arrays_slices.rs

PR #487, opened 2024-03-26
Add support for slice patterns with both prefix and suffix paterns

olympichek:
Previously `coq-of-rust` was able to translate only slice patterns with prefix patterns. This pull request implements support for slice patterns with suffix patterns as well. This fixes an issue (#270) with [match_destructuring_arrays_slices.rs](https://github.com/formal-land/coq-of-rust/blob/main/examples/rust_book/flow_of_control/match_destructuring_arrays_slices.rs) example.

* On Coq side, a new operation on slices (`get_slice_rev_index_or_break_match`) is added to `M.v`. Additionally, existing `get_slice_rest_or_break_match` operation is adjusted for better support of suffix patterns. I added comments with descriptions of intended semantics of these operations.
* On Rust side, struct that represents slice patterns now contains both prefix patterns and suffix patterns (previously it only had initial patterns). The code that uses it is adjusted to work with this new representation. Slice patterns with suffix patterns are translated to new operations on slices from `M.v`.


lib/src/thir_pattern.rs
```diff
@@ -178,31 +178,18 @@ pub(crate) fn compile_pattern(env: &Env, pat: &Pat) -> Rc<Pattern> {
                 suffix.iter().map(|pat| compile_pattern(env, pat)).collect();
             match slice {
```
clarus, line 179: We could do something like:
```rs
let slice_pattern = slice.map(|slice| compile_pattern(env, pat_middle));

Rc::new(Pattern::Slice {
    prefix_patterns: prefix,
    slice_pattern,
    suffix_patterns: suffix,
})
```

olympichek: Yea something like this should work

clarus (approved):

Issue #497, opened 2024-03-29
Maximum line size of 100


clarus:
Currently, the maximum line size in the generated Coq code is 80 characters. Make it 100. Reasons:

- in some cases, the code is deeply nested, so that would avoid some extreme cases where everything is split because the indentation itself reaches 80 characters;
- this is the default limit for Rust Format, so Rust programmers are probably more used to this size.

Issue #477, opened 2024-03-21
Uniformize warnings


clarus:
In https://github.com/formal-land/coq-of-rust/blob/main/lib/src/env.rs there is the function `emit_warning_with_note` to show a warning attached to a code snippet on the terminal.

The goal of this task is to use this function everywhere so that all warnings are shown in the same, always with an additional note. The note should either be a suggestion to avoid the warning or a message telling that we will take care of the issue / ask to report the issue. Lone warnings in the code can be found looking for `struct_span_warn`.

PR #501, opened 2024-04-01
Uniformize Warnings & 100 Character Line Width 

InfiniteEchoes:
This PR is for #477 and #497 , since the change to fix them are pretty small.
- Made the `note` for `emit_warning_with_note` optional
- Regarded other minor changes
- Adjusted line width to 100


lib/src/env.rs
```diff
@@ -12,11 +12,14 @@ pub(crate) fn emit_warning_with_note(
     env: &Env,
     span: &rustc_span::Span,
     warning_msg: &str,
-    note_msg: &str,
+    note_msg: &Option<&str>,
```
clarus, line 15:
```suggestion
    note_msg: Option<&str>,
```

lib/src/env.rs
```diff
@@ -12,11 +12,14 @@ pub(crate) fn emit_warning_with_note(
...
 ) {
-    env.tcx
+    let mut warn = env
```
clarus, line 17:
```suggestion
    let warn = env
```

InfiniteEchoes: If I remove the `mut`, the compiler would report that it cannot be borrowed as mutable in the match branches. Is there a way to solve this issue or I can just keep the `mut`?

clarus: OK yes I did not know, so we keep the `mut` then!

clarus (approved):

clarus:
OK, we can merge once the CI is green!

PR #502, opened 2024-04-01
New monadic notation

olympichek:
### Implement translation of the Rust programs using new monadic notation (#468)
This is an updated version of #469, made compatible with recent changes in the handling of types.

This pull request implements the following updates:
* `mt_expression()` function that was used to perform monadic transformation of expressions is removed as it is no longer needed with the representation using new notation
* Generated Coq code for bodies of functions, `const` items and lambda-expressions is wrapped in a call to the `M.monadic` tactic
* Monadic operations and function calls that are marked as effectful use new notation: `e (| e1 , .. , en |)`
* `M.monadic` tactic is significantly improved since its initial implementation
* All examples work, snapshots are updated. `MonadicNotationExample.v` removes as it is no longer needed


lib/src/coq.rs
```diff
@@ -407,6 +408,36 @@ impl<'a> Expression<'a> {
                     ),
                 ]),
             ),
+            Self::MonadicApplication { func, args } => paren(
+                with_paren,
+                nest([
+                    func.to_doc(false),
+                    line(),
```
clarus, line 415:
```suggestion
                    text(" "),
```
as I feel it is better to keep the opening parenthesis on the same line.

lib/src/coq.rs
```diff
@@ -407,6 +408,36 @@ impl<'a> Expression<'a> {
...
+                            nest([
+                                line(),
+                                intersperse(
+                                    args.iter().map(|(param, arg)| match param {
+                                        Some(param) => render::round_brackets(group([
+                                            text(param.to_owned()),
+                                            text(" := "),
+                                            arg.to_doc(false),
+                                        ])),
+                                        None => arg.to_doc(true),
```
clarus, line 430:
```suggestion
                                        None => arg.to_doc(false),
```
I think parenthesis are not necessary here (I might be wrong).

lib/src/coq.rs
olympichek: Ok

lib/src/coq.rs
olympichek: Yea they are indeed redundant. I'll remove them

clarus (approved):

Issue #495, opened 2024-03-29
Remove the StructUnit type


clarus:
In `Expr` there are two constructors:
```rust
    StructTuple {
        path: Path,
        fields: Vec<Rc<Expr>>,
    },
    StructUnit {
        path: Path,
    },
```

The second one could be viewed as a particular case for the first one, with `fields` equal to the empty vector. The goal of this task is to remove `StructUnit` using a `StructTuple` with empty `fields` instead.

Issue #478, opened 2024-03-21
[meta] Use the coq::Expression type to print expressions


clarus:
For now, most of the pretty printing of the expressions in https://github.com/formal-land/coq-of-rust/blob/main/lib/src/expression.rs is done using combinators from `render.rs`. We should instead use combinators from the `coq.rs` file, and replace `to_doc` functions by `to_coq` versions. Do the same for `top_level.rs` too, or all other area where we use `render.rs` instead of `coq.rs`.

The goal is to uniformize the code for the printing.

InfiniteEchoes:
There're 3 places I think would be unable to remove the `render.rs` dependency:
- `path.rs`: `Path` is actually a more low-level type in that `coq::Expression::just_name` depends on it. So its `to_doc` cannot be rewritten with `coq`'s combinators.
- `thir_expression.rs` and `thir_pattern.rs`: the only dependency that they are relying on is `render::capitalize()`. Although this function doesn't involve any dependencies with the rest of the code, there doesn't seem to have a better place to place this function other than `render.rs`.

PR #506, opened 2024-04-01
Misc Code Cleans(Updated)

InfiniteEchoes:
This PR is a updated version of #505 since there are too much conflicts to resolve with regard to the newest main branch.

This PR includes the following updates wrt several PRs:
- Removed StructUnit(#495 )
- Updated `ImplItemKind::to_coq` and `TopLevelItem::Comment::to_coq`
- With these updates, removed the `render` dependency for `top_level.rs` (#478 ) and all current occurences of `TopLevelItem::Code` & `coq::Expression::Code`.

In particular, `TopLevelItem::Comment` now supports multi-line comments. The comment style is fixed to
```coq
(* Single line comment regardless of the content *)
```
and
```coq
(*
Multi line
    comment
 regardless of the 
    content
*)
```

In my opinion the issues being mentioned can be closed after this PR.


lib/src/coq.rs
```diff
@@ -18,8 +18,9 @@ pub(crate) struct TopLevel<'a> {
 pub(crate) enum TopLevelItem<'a> {
     /// the Code variant is for those constructions
     /// that are not yet represented by the types in this file
+    #[allow(dead_code)]
     Code(Doc<'a>),
```
clarus, line 22: OK top that `Code` is not used anymore! We can also remove this constructor.

lib/src/coq.rs
```diff
@@ -70,6 +71,7 @@ pub(crate) enum DefinitionKind<'a> {
 pub(crate) enum Expression<'a> {
     /// the Code variant is for those constructions
     /// that are not yet represented by the types in this file
+    #[allow(dead_code)]
     Code(Doc<'a>),
```
clarus, line 75: Same, we can now remove this `Code` constructor.

InfiniteEchoes: Removing `<'a>` for `Expression` would produce a massive chain effect, letting the compiler to let you delete almost all lifetime params for almost all structs being defined in this file. I wonder if this is ok for you?

clarus: I think it is fine to do it in this pull request, and the opportunity to remove something that was making a lot of the code more complex.

InfiniteEchoes: Ok. The newest PR already removed them all

clarus (approved):

PR #523, opened 2024-04-08
Match pattern of `str` constants on `thir` level

InfiniteEchoes:


CoqOfRust/examples/default/examples/rust_book/error_handling/option_and_unwrap.v
```diff
@@ -28,6 +28,7 @@ Definition give_adult (τ : list Ty.t) (α : list Value.t) : M :=
                     "core::option::Option::Some",
                     0
                   |) in
+                let _ := M.is_constant_or_break_match (| M.read (| γ0_0 |), mk_str "lemonade" |) in
```
clarus, line 31:
```suggestion
                let _ := M.is_constant_or_break_match (| M.read (| γ0_0 |), Value.String "lemonade" |) in
```
as `mk_str` wraps a string in an address. I guess that here we want the string itself.

lib/src/thir_pattern.rs
```diff
@@ -100,7 +100,21 @@ pub(crate) fn compile_pattern(env: &Env, pat: &Pat) -> Rc<Pattern> {
         PatKind::Constant { value } => {
             if let rustc_middle::mir::Const::Ty(constant) = value {
                 let ty = constant.ty();
-
+                // Brutal way to handle the case of rustc_middle::ty::TyKind::Str
+                // Since the type would be erased when it comes down to thir level
+                let kind_name = format!("{:?}", ty.kind());
+                if kind_name == *"&ReErased str".to_string() {
+                    let string_value = format!("{}", constant);
```
clarus, line 107:
```suggestion
                    let string_value = constant.to_string();
```

lib/src/thir_pattern.rs
```diff
@@ -100,7 +100,21 @@ pub(crate) fn compile_pattern(env: &Env, pat: &Pat) -> Rc<Pattern> {
         PatKind::Constant { value } => {
             if let rustc_middle::mir::Const::Ty(constant) = value {
                 let ty = constant.ty();
-
+                // Brutal way to handle the case of rustc_middle::ty::TyKind::Str
+                // Since the type would be erased when it comes down to thir level
```
clarus, line 104:
```suggestion
                // Since the type would be erased when it comes down to THIR level
                // TODO: have a translation that works for all strings
```
I guess that for strings that contain special characters such as `"` we would have a limitation.

InfiniteEchoes: I believe that the trim should work fine. There would always have an extra `""` pair wrapped around the original string, and I trimmed them out. The rest of the content are being processed by `string_pieces_to_coq` which handles such case.

PR #524, opened 2024-04-08
Add mutual dependency loop example

InfiniteEchoes:
A custom example code that would bypass the compiler and cause the program to crash.


clarus:
Merging https://github.com/formal-land/coq-of-rust/pull/524 instead cherry-picking the commit with the new example

PR #540, opened 2024-05-01
Experimentation with Solana Token-2022 program

olympichek:
An attempt to translate Solana Token-2022 program with coq-of-rust and compile translation results with Coq.
- [CoqOfRust/token-2022](https://github.com/formal-land/coq-of-rust/tree/valerii-huhnin%40token-2022-compile-experiment/CoqOfRust/token-2022) contains translation results

All files compile now. Two changes were made to make it work:
- All `const _ = ...` definitions are now skipped during translation to avoid name collision problems, see [9ce7b1d](https://github.com/formal-land/coq-of-rust/pull/540/commits/9ce7b1d05a6cbac19c9a0b0560a552914819b27e)
- New monadic operation `M.yield` was added  (for now, as a parameter) to support this Rust feature, see [2012ea2](https://github.com/formal-land/coq-of-rust/pull/540/commits/2012ea2c73074f0c42b48e40ba3ce5dbb9e4fb9b)


lib/src/top_level.rs
```diff
@@ -334,8 +334,8 @@ fn compile_top_level_item_without_local_items<'a>(
             if check_if_test_declaration(ty) {
                 return vec![];
             }
-            // skip const _ : () = ...
-            if item.ident.name.as_str() == "_" && matches!(ty.kind, TyKind::Tup([])) {
+            // skip const _ : ... = ...
```
clarus, line 337: We can leave it there for now, but I wonder if we could miss some side effects, like in initialization functions. Anyway, a good practice is to have an explicit `init` function instead of things that get executed when a module is loaded.

clarus (commented):
We can move the `token-2022` folder to `CoqOfRust/` as `examples/default` are for the single-file examples.

lib/src/top_level.rs
olympichek: So do you thing there might be examples where those "underscore" definitions are somehow used/called or do you mean that they can be used somehow indirectly due to side effects? If this is the latter, then how is it even handled within our semantics?

clarus: Yes, you are right, we do not handle side effects for `const` even when they are named! I have not yet seen code where this is used, so I think we can remove the `const _` for now, and if in the future we encounter some code relying on side effects in `const`, we will think about what to do.

By "We can leave it there for now" I meant that we can keep your change, in case this was unclear.

olympichek:
> We can move the `token-2022` folder to `CoqOfRust/` as `examples/default` are for the single-file examples.

Done

clarus (approved):

PR #552, opened 2024-05-26
Experiment: linking

clarus:
See https://github.com/formal-land/coq-of-rust/pull/556 for the cleaned version of this pull request

The goal is to have a first step where we:

- add the types
- make name resolution (including trait resolution)

These steps are tedious and can probably be done once and for all in a separate step, as here.

TODO:

- add some doc
- clean the code of the pull request
- automatically infer the `to_value` for read and write operations, from the reference


clarus:
See https://github.com/formal-land/coq-of-rust/pull/556 for the new PR

PR #553, opened 2024-05-27
Add two simulation files for rust example code

InfiniteEchoes:


clarus:
We are currently changing the way we do proofs to be able to handle larger Rust codebases. So we will do these simulations later. A current prospect for updates is https://github.com/formal-land/coq-of-rust/pull/552

PR #579, opened 2024-06-28
Various updates of monads and lenses for simulations

olympichek:
This pull request implements various updates of monads and lenses for simulations:
- Instead of the `Error` monad which was used previously, `Panic` and `Result` monads are defined to model different kinds of error handling within Rust programs. To complement the `Result` monad, `Option` monad is also defined
- New notations for these monads: `M?` for `Option` monad, `M??` for `Result` monad and `M!?` for `Panic` monad
- New lens: `LensOption`. For this lens, `read` and `write` operations can fail by returning `None` value. The `lift` function for this lens allows to lift state monad computation that returns `option` value
- New lens: `LensPanic`. For this lens, `read` and `write` operations can fail with a panic. If within the `lift` function one of these operations fail, than the result would be the computation that panics
- Notations for different structures are now defined in separate modules, but single `Notations` module that re-exports all notations is also added

Additionally, REVM stack operations are rewritten using `LensPanic`. This provides an example of how lenses can be used to represent subpointers (mutable references). In this example, `LensPanic` is necessary, as the lens is used to represent `top_unsafe` operation, which fails with a panic when called on an empty stack.

### DIFF
--- a/CoqOfRust/revm/simulations/interpreter/instructions/arithmetic.v
+++ b/CoqOfRust/revm/simulations/interpreter/instructions/arithmetic.v
@@ -26,9 +26,12 @@ Require Import CoqOfRust.revm.simulations.interpreter.instructions.i256.
 Definition add :
     MS? Interpreter.t string unit :=
   letS? _ := gas_macro Gas.VERYLOW in
-  letS? '(op1, op2) := pop_top_macro2 in
+  letS? '(op1, op2_ref) := pop_top_macro2 in
   liftS? Interpreter.Lens.stack (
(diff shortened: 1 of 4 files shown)


InfiniteEchoes:
Is it that the `Result` and `Option` monads are corresponded exactly to the `Result` & `Option` type in Rust?

olympichek:
> Is it that the `Result` and `Option` monads are corresponded exactly to the `Result` & `Option` type in Rust?

Yes, that's the idea

clarus (approved):

PR #583, opened 2024-07-23
Move `AbstractStack` Tests

olympichek:
This pull request adds simulations for tests for Move `AbstractStack` methods.

Additionally, this pull request adds:
- simulations for `Eq` Trait and its implementation for some of the core types and Move's `AbsStackError` type
- simulations for Rust's `assert` and `assert_eq` macros, that are used to write simulations for tests
- `foldS?` operation: similar to `List.fold_left` but for functions that return a value wrapped in a monad
- various improvements and fixes for simulations of `AbstractStack` methods

In Rust, tests use `assert` and `assert_eq` macros, which cause a panic if the test condition fails. We can write simulations for Rust tests the same way we write simulations for normal Rust functions. Then, we can state a lemma that simulation passes the test, i.e. that an execution of the simulation does not result with a panic. Such lemma can be trivially proven with computation and application of `reflexivity`. By ensuring that our simulations pass the same tests that the Rust code passes, i.e. by ensuring that our simulations and the Rust code behave equivalently on the provided test cases, we can argue that our simulations behave equivalently to the Rust code in general.

### DIFF
--- a/CoqOfRust/move_sui/simulations/move_abstract_stack/lib.v
+++ b/CoqOfRust/move_sui/simulations/move_abstract_stack/lib.v
@@ -197,7 +197,7 @@ Module AbstractStack.
     }
   *)
 
-  Definition pop_eq_n {A : Set} (item : A) (n : Z) :
+  Definition pop_eq_n {A : Set} (n : Z) :
       MS? (t A) string (Result.t A AbsStackError.t) :=
     letS? self := readS? in
     if (is_empty self || (n >? len self))%bool 
@@ -225,9 +225,9 @@ Module AbstractStack.
     }
   *)
 
-  Definition pop {A : Set} (item : A) :
+  Definition pop {A : Set} :
       MS? (t A) string (Result.t A AbsStackError.t) :=
-    pop_eq_n item 1.
+    pop_eq_n 1.
 
   (*
     /// Pop any n items off the stack. Unlike `pop_n`, items do not have to be equal
--- a/CoqOfRust/move_sui/simulations/move_abstract_stack/lib.v
+++ b/CoqOfRust/move_sui/simulations/move_abstract_stack/lib.v
@@ -4,8 +4,10 @@ Require Import CoqOfRust.core.simulations.integers.
 Require Import CoqOfRust.core.simulations.vector.
 Require Import CoqOfRust.core.simulations.option.
 Require Import CoqOfRust.core.simulations.eq.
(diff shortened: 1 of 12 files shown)


CoqOfRust/move_sui/simulations/move_abstract_stack/unit_tests/mod.v
```diff
@@ -0,0 +1,57 @@
...
+  letS? _ :=
+    assert_eqS?
+      (AbstractStack.pop_eq_n 112)
+      (returnS? (Result.Err AbsStackError.Underflow)) in
+  assertS? (AbstractStack.self_is_empty).
+
+Lemma test_empty_stack_correct :
+  testS? test_empty_stack AbstractStack.new.
+Proof.
+  reflexivity.
```
clarus, line 56: Maybe later, we will need to add a `vm_compure` here to speed up the tests if we get large examples!

clarus (approved):

CoqOfRust/move_sui/simulations/move_abstract_stack/lib.v
```diff
@@ -19,6 +21,30 @@ Module AbsStackError.
   | ElementNotEqual
   | Underflow
   | Overflow.
+
+  Module ImplEq.
+    Global Instance I :
+      Eq.Trait AbsStackError.t := {
+        eqb e1 e2 :=
+          match e1 with
+          | AbsStackError.ElementNotEqual =>
```
InfiniteEchoes, line 30: I have a feeling that whenever there're boolean equality predicates using on constructors, `Scheme Equality` of Coq could come in handy...

olympichek: Yea, this is indeed quite handy. I didn't know that such feature exists, thank you for a suggestion!

olympichek: What is the difference between `Scheme Equality` and `Scheme Boolean Equality` though?

InfiniteEchoes: When I use `Scheme Boolean Equality` it generates a function that returns a bool. I haven't looked into other situations yet so for now I could only guess - I'm guessing that `Scheme Equality` might return a `Prop` Instead

clarus (approved):

clarus:
OK top!

PR #584, opened 2024-07-24
More on `type_safety`, part 3

InfiniteEchoes:
Plans or features:
- [ ] Implement `CompiledModule`
- [x] Implement `Scope` for `Meter` trait
- [x] Implement `Bounds` for `Meter` trait
- [x] Implement the functions in `Meter` trait and specifically two structs that implements this trait
- [ ] Implement `verify_instr`
- [x] Import `AbstractStack` into `type_safety` and used it correctly
- [ ] to be updated...

### DIFF
--- a/CoqOfRust/move_sui/simulations/move_bytecode_verifier_meter/lib.v
+++ b/CoqOfRust/move_sui/simulations/move_bytecode_verifier_meter/lib.v
@@ -4,6 +4,183 @@ Require Import CoqOfRust.lib.lib.
 
 Import simulations.M.Notations.
 
+Require CoqOfRust.move_sui.simulations.move_binary_format.errors.
+Module PartialVMResult := errors.PartialVMResult.
+Module PartialVMError := errors.PartialVMError.
+
+(* TODO(progress):
+- Implement `BoundMeter` just in case
+- Write out the exact function chains from `verify_instr`
+- Implement `Scope` or stub it
+- Maybe move `BoundMeter` and `DummyMeter` to their files
+*)
+
+(* 
+use crate::{Meter, Scope};
+*)
+(* pub struct DummyMeter; *)
+Module DummyMeter.
+  Record t : Set := { }.
+
+  (* impl Meter for DummyMeter  *)
+  Module Impl_move_sui_simulations_move_bytecode_verifier_meter_DummyMeter.
+    Definition Self : Set := 
+      move_sui.simulations.move_bytecode_verifier_meter.DummyMeter.t.
+
+    (* fn enter_scope(&mut self, _name: &str, _scope: Scope) {} *)
+    Definition enter_scope (self : Self) (_name : str) (_scope : Scope) := .
+
+    (* fn transfer(&mut self, _from: Scope, _to: Scope, _factor: f32) -> PartialVMResult<()> { Ok(()) } *)
+    (* TODO: translate the `f32` here correctly *)
+    Definition transfer (self : Self) (_from : Scope) (_to : Scope) (_factor : Z) : PartialVMResult.t unit := return?? tt.
+
+    (* fn add(&mut self, _scope: Scope, _units: u128) -> PartialVMResult<()> { Ok(()) } *)
+    Definition add (self : Self) (_scope : Scope) (_units : Z) : PartialVMResult.t unit := return?? tt.
+  End Impl_move_sui_simulations_move_bytecode_verifier_meter_DummyMeter.
+End DummyMeter.
+
+(* bound.rs
+// Copyright (c) The Move Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{Meter, Scope};
+use move_binary_format::errors::{PartialVMError, PartialVMResult};
+use move_core_types::vm_status::StatusCode;
+use move_vm_config::verifier::MeterConfig;
+
+/// Module and function level metering.
+
+impl Bounds {
+    fn add(&mut self, units: u128) -> PartialVMResult<()> {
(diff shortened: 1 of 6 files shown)


clarus (approved):

CoqOfRust/move_sui/simulations/move_bytecode_verifier_meter/lib.v
```diff
@@ -4,7 +4,347 @@ Require Import CoqOfRust.lib.lib.
...
+          }
+      }
+      *)
+      Definition get_bounds_mut (self : Self) (scope : Scope.t) : M!? string Bounds.t :=
+        match scope with
+        | Scope.Package     => return!? self.(pkg_bounds)
+        | Scope.Module      => return!? self.(mod_bounds)
+        | Scope.Function    => return!? self.(fun_bounds)
+        | Scope.Transaction => panic!? "transaction scope unsupported."
+        end.
```
olympichek, line 295: Here we have a function that takes a mutable reference and returns a mutable reference to its component. Such functions are good candidates to be represented using a lens. As it can also panic we can use `LensPanic` here

olympichek: I think something like this would work:
```Coq
Definition get_bounds_mut (scope : Scope.t) :
    LensPanic.t string State Bounds.State := {|
  LensPanic.read boundmeter := 
    match scope with
    | Scope.Package     => return!? (boundmeter.(pkg_bounds))
    | Scope.Module      => return!? (boundmeter.(mod_bounds))
    | Scope.Function    => return!? (boundmeter.(fun_bounds))
    | Scope.Transaction => panic!? "transaction scope unsupported."
     end;
  LensPanic.write boundmeter bounds := 
    match scope with
    | Scope.Package     => return!? (boundmeter <| pkg_bounds := bounds |>)
    | Scope.Module      => return!? (boundmeter <| mod_bounds := bounds |>)
    | Scope.Function    => return!? (boundmeter <| fun_bounds := bounds |>)
    | Scope.Transaction => panic!? "transaction scope unsupported."
     end;
|}.
```

CoqOfRust/move_sui/simulations/move_bytecode_verifier_meter/lib.v
```diff
@@ -4,7 +4,347 @@ Require Import CoqOfRust.lib.lib.
...
+        let bounds := bounds <| Bounds.units := 0    |> in 
+        let self := match scope with 
+          | Scope.Package     => self <| BoundMeter.pkg_bounds := bounds |> 
+          | Scope.Module      => self <| BoundMeter.mod_bounds := bounds |> 
+          | Scope.Function    => self <| BoundMeter.fun_bounds := bounds |>
+          | Scope.Transaction => self
+        end in
+        let state : State := {| self := self |} in
+          writeS? state.
+
```
olympichek, line 316: If we define `get_bounds_mut` as a `LensPanic` like I suggested above, we can defeine `enter_scope` using it like this:
```Coq
Definition enter_scope (name : string) (scope : Scope.t) : MS? State string unit :=
  liftS?of!? (get_bounds_mut scope) (
    letS? bounds := readS? in
    letS? _ := writeS? (bounds <| Bounds.name  := name |>) in 
    letS? _ := writeS? (bounds <| Bounds.units := 0    |>) in
    returnS? tt
  ).
```

CoqOfRust/move_sui/simulations/move_bytecode_verifier_meter/lib.v
```diff
@@ -4,7 +4,347 @@ Require Import CoqOfRust.lib.lib.
...
+      *)
+      Definition add (self : Self) (scope : Scope.t) (units : Z) 
+        : MS? State string (PartialVMResult.t unit) :=
+        letS? _ := writeS? (Build_State self) in
+        letS? bounds := return!?toS? (get_bounds_mut self scope) in
+        letS? boundmeter := liftS? 
+          (Lens_BoundMeter_State_Bounds_State.values scope) 
+          (Bounds.Impl_move_sui_simulations_move_bytecode_verifier_meter_Bounds.add 
+          bounds units)
+          in
```
olympichek, line 330: Then here we can use `get_bounds_mut` lens directly

CoqOfRust/move_sui/simulations/move_bytecode_verifier_meter/lib.v
```diff
@@ -4,7 +4,347 @@ Require Import CoqOfRust.lib.lib.
...
+        (* NOTE: is it the correct way to write the updated value? *)
+        letS? self := readS? in
+        letS? _ := writeS? self in
```
olympichek, line 333: I am pretty sure that performing `writeS?` operation directly after `readS?` operation with exactly the value that `readS?` returned is just equivalent to identity function, so it does nothing basically. So basically I think that you can just omit this two lines - invocation of the `liftS?` should be enough, you don't need to synchronize the state after it or something

CoqOfRust/simulations/M.v
```diff
@@ -116,6 +116,8 @@ Module StatePanic.
   Definition write {State Error : Set} (state : State) : t State Error unit :=
     fun _ => (return!? tt, state).
 
+  (* TODO: change the `string` to arbitary type `Error` *)
+  (* I think this combinator might be heavily bugged *)
```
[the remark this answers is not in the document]
InfiniteEchoes: > Is something particularly wrong with it?

No, this should be a deprecated comment, as we have discussed in the Discord.

olympichek (approved):

Issue #575, opened 2024-06-20
Move - Simulations - Verifier - verify_instr function


clarus:
Write a simulation for the function `verify_instr` in https://github.com/move-language/move-sui/blob/main/crates/move-bytecode-verifier/src/type_safety.rs

PR #590, opened 2024-08-12
`verify_instr` simulation

InfiniteEchoes:
After #587 , pretty much of the dependencies have been resolved nicely.

This PR is focusing on the main task: #575 .

This PR:
- [x] Resolve mutual dependency issue
- [x] Translate the `verify_instr` completely

# Known Issues
1. `Err` value might not be propagated correctly and cannot be detected by the type checker
2. `safe_unwrap_err` might not work correctly. They're intended to be fixed together.
3. Roughly seeing so far, the `verify` function seem to need much more dependency to implement, so I haven't implement it
4. In current implementation, we ignore any messages for `PartialVMError` for simplicity


CoqOfRust/move_sui/simulations/move_bytecode_verifier/type_safety.v
```diff
@@ -465,8 +674,59 @@ fn call(
...
+  (type_actuals : Signature.t) 
+  : MS? (TypeSafetyChecker.t * Meter.t) string (PartialVMResult.t unit) :=
+  letS? '(verifier, _) := readS? in
+  let parameters := CompiledModule
+    .Impl_move_sui_simulations_move_binary_format_file_format_CompiledModule
+    .signature_at verifier.(TypeSafetyChecker.module) 
+      function_handle.(FunctionHandle.parameters) in
+  let parameters := List.rev parameters.(Signature.a0) in
+  let fold :=
+  (fix fold (l : list SignatureToken.t)
```
olympichek, line 687: I see that in several places you use the internally defined `fold` function. Is there some general pattern here? Perhaps something like `List.fold_left` or `foldS?` (a monadic variant of it that I defined recently for simulations) could be used?

CoqOfRust/move_sui/simulations/move_bytecode_verifier/type_safety.v
```diff
@@ -373,11 +481,63 @@ fn borrow_field(
...
+    letS? operand := liftS? TypeSafetyChecker.lens_self_meter_self (
+      liftS? TypeSafetyChecker.lens_self_stack AbstractStack.pop) in
+    letS? operand := safe_unwrap_err operand in
+
+    if andb mut_ (negb $ SignatureToken
+      .Impl_move_sui_simulations_move_binary_format_file_format_SignatureToken
+      .is_mutable_reference operand)
+    then returnS? $ Result.Err $ 
+      TypeSafetyChecker
+      .Impl_move_sui_simulations_move_bytecode_verifier_type_safety_TypeSafetyChecker
```
olympichek, line 497: This is more of a style question, but is it necessary for us to have such long names for some functions, like `Impl_move_sui_simulations_move_bytecode_verifier_type_safety_TypeSafetyChecker` ? Especially considering that we use modules already and in the case of ambiguity we can use a qualified name (which includes the name of the module) ?

CoqOfRust/move_sui/simulations/move_bytecode_verifier/type_safety.v
InfiniteEchoes: > I see that in several places you use the internally defined `fold` function. Is there some general pattern here? Perhaps something like `List.fold_left` or `foldS?` (a monadic variant of it that I defined recently for simulations) could be used?

It is just because I want to simulate the `for` loops in the Rust code and I didn't want to do the brain work to figure out how to present it using the Coq's built-in fold. I think the pattern I can describe should be
- `fold` takes in the `list` to `iter`ate on the original `for` loop. There might be other parameters either saved in the states of the monad or being provided additionally
- In the body of `fold` we simulate the original for loop's body until the `list` is empty. 

In short: I don't notice anything very special about these `fold` loops on monad, they're pretty much like normal folds.

And yes, if we have a `foldS?` that definitely helps a lot.

olympichek: Well, okay

olympichek (approved):

InfiniteEchoes:
I'm done with everything that I need to do.

Issue #496, opened 2024-03-29
Handle const polymorphic parameters


clarus:
This should help translating the `core` crate that is currently missing this feature to translate.

PR #604, opened 2024-08-26
Add handling of polymorphic constants

clarus:
Here is an example of polymorphic constant in Rust:
```rust
fn print_array<T: std::fmt::Debug, const N: usize>(arr: [T; N]) {
    println!("{:?}", arr);
}

fn main() {
    let arr1 = [1, 2, 3];
    let arr2 = [4, 5, 6, 7, 8];

    print_array(arr1);
    print_array(arr2);

    let char_arr: [char; 3] = ['a', 'b', 'c'];
    print_array(char_arr);
}
```
Here, the polymorphic constant is `N`. In this pull request, we add such parameters with the same handling as for polymorphic types, which are also represented as values in Coq.

The aim is to be able to compile the Move Sui project that depends on such a feature to compile.


clarus:
Fixes https://github.com/formal-land/coq-of-rust/issues/496

PR #622, opened 2024-11-13
Add ReadRef & WriteRef implementation to the interpreter.

Mushow:


CoqOfRust/move_sui/simulations/move_vm_runtime/interpreter.v
```diff
@@ -1053,7 +1079,13 @@ Definition execute_instruction (pc : Z)
...
   }
   *)
-  | Bytecode.WriteRef => returnS! $ Result.Ok InstrRet.Ok
+  | Bytecode.WriteRef =>
+  letS!? reference := liftS! Interpreter.Lens.lens_state_self (
+    liftS! Interpreter.Lens.lens_operand_stack $ Stack.Impl_Stack.pop_as Reference.t) in
+  letS!? value := liftS! Interpreter.Lens.lens_state_self (
+    liftS! Interpreter.Lens.lens_operand_stack Stack.Impl_Stack.pop) in
+  letS!? _ := returnS! $ Impl_ReferenceImpl.write_ref reference value in
+  returnS! $ Result.Ok InstrRet.Ok
```
clarus, line 1088: With one more level of indentation

CoqOfRust/move_sui/simulations/move_vm_types/values/values_impl.v
```diff
@@ -1,4 +1,6 @@
 Require Import CoqOfRust.CoqOfRust.
+Require Import Coq.Lists.List.
```
clarus, line 2: We should not need this line, as we prefix all the calls to list functions with `List.` to help to know where they are from when reading the code.

CoqOfRust/move_sui/simulations/move_vm_types/values/values_impl.v
```diff
@@ -1,4 +1,6 @@
...
+Import ListNotations.
```
clarus, line 3: These notations should already be in `Require Import CoqOfRust.CoqOfRust.`.

CoqOfRust/move_sui/simulations/move_vm_types/values/values_impl.v
```diff
@@ -7,6 +9,7 @@ Import simulations.M.Notations.
 Require CoqOfRust.move_sui.simulations.move_binary_format.errors.
 Module PartialVMResult := errors.PartialVMResult.
 Module PartialVMError := errors.PartialVMError.
+Import PartialVMResult.
```
clarus, line 12: I think we should avoid this `Import` otherwise it is too unclear where the calls are from.

CoqOfRust/move_sui/simulations/move_vm_types/values/values_impl.v
```diff
@@ -153,11 +203,13 @@ enum ReferenceImpl {
     ContainerRef(ContainerRef),
 }
 *)
+
 Module ReferenceImpl.
   Inductive t : Set :=
   | IndexedRef : IndexedRef.t -> t
   | ContainerRef : ContainerRef.t -> t
   .
+
```
clarus, line 212: I think we should not have an empty line at the end of modules actually, even if this is often the case in our current code.

Mushow:
I will put these changes to effect. What about the L. 687 in **values_impl.v** ?

clarus (approved):

PR #648, opened 2025-01-24
draft: implement eof_create_inputs definitions

0xMushow:


CoqOfRust/revm/links/interpreter/interpreter_action/eof_create_inputs.v
```diff
@@ -31,14 +86,78 @@ Module EOFCreateInput.
...
+            kind,
+        }
+    }
+  *)
+
+  Parameter address_default : Address.t.
+  Parameter eof_decode : Bytes.t -> Eof.t.
+
+  Definition new (caller : Address.t) (value : U256.t)
+                 (gas_limit : U64.t) (kind : EOFCreateKind.t) : Self :=
```
clarus, line 120: We do not create any definitions in the links mode and only write `run_new`. This is one advantage of the links approach. Here the statement would be something like:

```coq
Definition run_new (caller : Address.t) (value : U256.t) (gas_limit : U64.t) (kind : EOFCreateKind.t) :
  {{
    (* the name is in the translation file and might be longer with more module prefixes *)
    Impl_revm_interpreter_interpreter_action_eof_create_inputs_EOFCreateInputs.new
      [] [] [φ caller; φ value; φ gas_limit; φ kind] ⇓
    fun (v : Self) => inl (φ v)
  }}.
Proof.
```

Then for an example of proof there is the file https://github.com/formal-land/coq-of-rust/blob/main/CoqOfRust/revm/links/interpreter/interpreter/gas.v , although it is still quite experimental.

CoqOfRust/revm/links/interpreter/interpreter_action/eof_create_inputs.v
```diff
@@ -31,14 +86,78 @@ Module EOFCreateInput.
...
+                created_address,
+            },
+        )
+    }
+  *)
+
+  Definition new_opcode (caller : Address.t) (created_address : Address.t)
+                        (value : U256.t) (eof_init_code : Eof.t)
+                        (gas_limit : U64.t) (input : Bytes.t) : Self :=
+    new caller value gas_limit (EOFCreateKind.Opcode eof_init_code input created_address).
```
clarus, line 160: Same as with the `new` function!

PR #681, opened 2025-03-06
Update the Rust version to 2024-12-07 + fix trait generics in impl

clarus:


clarus:
There are some constants that disappeared in the patterns that I need to investigate...

clarus:
> There are some constants that disappeared in the patterns that I need to investigate...

It should be fixed with the last commit.

PR #684, opened 2025-03-11
Link file for `block_info.v`

InfiniteEchoes:
This PR:
- Completed several operations in `arithmetic.v` since it's a good source of reference
- Added several functions in `block_info.v`'s link file, to be finished in the next PR

Other todos in future PR:
- Finish `Into`'s link in ruint.links.modular, start from `convert/links/mod.v`
- Resolve confilcts in `instructions.v`

PR #686, opened 2025-03-12
Suggestion commits for review chain_id

clarus:


clarus:
This is a commit of suggestions for https://github.com/formal-land/coq-of-rust/pull/684

Issue #723, opened 2025-04-15
ZK: add a translation of Plonky3


clarus:
Add a translation of the Plonky3 library https://github.com/Plonky3/Plonky3 into the `CoqOfRust/` folder, in the same way for example `revm/` or the `core/` library` are translated.

This implies:

- To add a Git submodule in the `third-party` folder with a specific commit of Plonky3 (we can take the last current one if it works).
- Updating https://github.com/formal-land/coq-of-rust/blob/main/.github/workflows/rust.yml to also have the translation of Plonky3.
- Making sure that the translated files are compiling. If this is not the case, we can add the files that do not work in https://github.com/formal-land/coq-of-rust/blob/main/CoqOfRust/blacklist.txt We can fix these files later if there are too many broken files.

Hopefully, the Plonky3 project will compile with the Rust version used by `coq-of-rust`. If this is not the case, we can take an older commit of Plonky3 so that it works (it worked at some point a few months ago).

clarus:
I am closing the issue as it is done!

PR #725, opened 2025-04-16
Plonky3 Translation

InfiniteEchoes:
For #723.

--------

Notes on special cases and blacklisted files:

**`util/src/lib.rs`**
We have to add a line of `#![feature(const_swap)]` at the beginning to compile the file..

**`plonky3/commit/src/adapters/extension_mmcs.v`**
**`plonky3/merkle-tree/src/hiding_mmcs.v`**
**`plonky3/merkle-tree/src/mmcs.v`**
Coq generates the following error:
```
Error: The reference M_ was not found in the current environment.
```

**`plonky3/monty-31/src/monty_31.v`**
```
line 649, characters 22-27:
Error:
In environment
MP : Ty.t
ε : list Value.t
τ : list Ty.t
α : list Value.t
Self := Impl_p3_monty_31_monty_31_MontyField31_MP.Self MP : Ty.t
N : Value.t
l : list Value.t
M_ : Value.t
l0 : list Value.t
input : Value.t
l1 : list Value.t
input0 := alloc (|input|) : Value.t
The term "[M]" has type "list Set" while it is expected to have type
 "list Value.t".
```

**`plonky3/poseidon2/src/external.v`**
```
line 2897, characters 48-55:
Error: The variable initial was not found in the current environment.
```

----------------

Since build time has been longer than expected, `make_and_detect_warnings` has been increased its `make` waiting time from 1 hour to 2.


clarus:
I squashed the commits in https://github.com/formal-land/coq-of-rust/pull/727 with a reduced number of folders so that we do not add too many files in the project, for the compilation time, ...

If we need to add more again, we have this pull request to keep the history.

PR #727, opened 2025-04-17
feat: Add `plonky3` translation

clarus:
This takes the work in https://github.com/formal-land/coq-of-rust/pull/725 with fewer translated folders (only those that will be necessary for the verification of Blake3).


clarus:
I will merge once the CI is green

PR #738, opened 2025-05-19
Translating p-token

romefeller:
@clarus


clarus:
OK top! You can:

- Put what is in `src/` directly, to have paths like `CoqOfRust/p_token/entrypoint.v` instead of `CoqOfRust/p_token/src/entrypoint.v`
- Add the scripts to https://github.com/formal-land/coq-of-rust/blob/main/.github/workflows/rust.yml by doing a copy&paste of the other scripts using `rsync`

clarus:
Actually I am just merging and we can see that later, it should be easy to add after the links, etc...

PR #836, opened 2026-06-23
Update Revm v103 memory resize simulate

cascade77:
This PR continues the Revm v103 small compile-slice work after the memory expansion macro update.

It focuses on the memory resize helper path, adapting it to the current v103 structure where resizing goes through the shared-memory helper and the interpreter's direct gas field.

Local check:
- `make -k revm/revm_interpreter/instructions/simulate/system/memory_resize.vo`

The fake-body scan is clean. No local translated bodies or wrong-argument stubs were added.

There is one focused admitted `_eq` bridge for the shared-memory resize helper, kept at that helper boundary so this PR does not expand into the lower memory-gas proof work.

### DIFF
--- a/RocqOfRust/revm/revm_interpreter/instructions/simulate/macros.v
+++ b/RocqOfRust/revm/revm_interpreter/instructions/simulate/macros.v
@@ -57,6 +57,43 @@ Ltac require_non_staticcall_macro_eq InterpreterTypesEq :=
     s
   |].
 
+Definition halt {WIRE : Set} `{Link WIRE}
+    {WIRE_types : InterpreterTypes.Types.t} `{InterpreterTypes.Types.AreLinks WIRE_types}
+    {IInterpreterTypes : InterpreterTypes.C WIRE_types}
+    (interpreter : Interpreter.t WIRE WIRE_types)
+    (result : instruction_result.InstructionResult.t) :
+    Interpreter.t WIRE WIRE_types :=
+  let action :=
+    interpreter_action.InterpreterAction.Return {|
+      InterpreterResult.result := result;
+      InterpreterResult.output := Impl_Bytes.new;
+      InterpreterResult.gas := interpreter.(Interpreter.gas);
+    |} in
+  interpreter
+    <| Interpreter.bytecode :=
+      IInterpreterTypes
+        .(InterpreterTypes.LoopControl_for_Bytecode)
+        .(LoopControl.set_action)
+        interpreter.(Interpreter.bytecode)
+        action
+    |>.
+
+Definition halt_oog {WIRE : Set} `{Link WIRE}
+    {WIRE_types : InterpreterTypes.Types.t} `{InterpreterTypes.Types.AreLinks WIRE_types}
+    {IInterpreterTypes : InterpreterTypes.C WIRE_types}
+    (interpreter : Interpreter.t WIRE WIRE_types) :
+    Interpreter.t WIRE WIRE_types :=
+  halt
+    (interpreter <| Interpreter.gas := Impl_Gas.spend_all interpreter.(Interpreter.gas) |>)
+    instruction_result.InstructionResult.OutOfGas.
+
+Definition halt_memory_oog {WIRE : Set} `{Link WIRE}
+    {WIRE_types : InterpreterTypes.Types.t} `{InterpreterTypes.Types.AreLinks WIRE_types}
+    {IInterpreterTypes : InterpreterTypes.C WIRE_types}
+    (interpreter : Interpreter.t WIRE WIRE_types) :
+    Interpreter.t WIRE WIRE_types :=
+  halt interpreter instruction_result.InstructionResult.MemoryOOG.
+
 Definition gas_macro {WIRE K : Set} `{Link WIRE}
     {WIRE_types : InterpreterTypes.Types.t} `{InterpreterTypes.Types.AreLinks WIRE_types}
     {IInterpreterTypes : InterpreterTypes.C WIRE_types}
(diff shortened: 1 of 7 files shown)


RocqOfRust/revm/revm_interpreter/instructions/simulate/macros.v
```diff
@@ -57,6 +57,43 @@ Ltac require_non_staticcall_macro_eq InterpreterTypesEq :=
     s
   |].
 
+Definition halt {WIRE : Set} `{Link WIRE}
```
clarus: The definition for `halt` and similar functions/lemmas should be put in the simulate file corresponding to where it is defined, so the simulate file for `revm/revm_interpreter/interpreter.rs`.

RocqOfRust/revm/revm_interpreter/instructions/simulate/macros.v
```diff
@@ -333,6 +292,116 @@ Proof.
...
+  ] unfold Impl_Interpreter.run_halt_memory_oog, Impl_Interpreter.run_halt.
+  cbn.
+  repeat s.
+  - apply Impl_Bytes.new_eq.
+  - apply InterpreterTypesEq
+      .(InterpreterTypes.Eq.LoopControl_for_Bytecode)
+      .(LoopControl.BytecodeEq.set_action).
+  - repeat rewrite Stack.dealloc_alloc_eq;
+    repeat rewrite stack_dealloc_cons_alloc_unit;
+    reflexivity.
```
clarus: OK cool for the proof!

RocqOfRust/revm/revm_interpreter/interpreter/simulate/shared_memory.v
```diff
@@ -17,3 +23,98 @@ Proof.
   }
   s.
 Qed.
+
+Definition resize_memory_cold
```
clarus, line 27: Could there be an `_eq` lemma as well?

RocqOfRust/revm/revm_interpreter/interpreter/simulate/shared_memory.v
```diff
@@ -17,3 +23,98 @@ Proof.
...
+    SubPointer.Runner.apply
+      ref_interpreter.(Ref.core)
+      Interpreter.SubPointer.get_memory
+  |} in
+  let result :=
+    @resize_memory
+      WIRE_types.(InterpreterTypes.Types.Memory)
+      WIRE_types.(InterpreterTypes.Types.Memory_Synthetic)
+      WIRE_types.(InterpreterTypes.Types.Memory_Synthetic1)
+      H0.(InterpreterTypes.Types.H_Memory)
```
clarus: It seems some trait parameter are duplicated. This can typically block proofs after but it is hard to catch for some reasons. Using `!` in:

```coq
{IInterpreterTypes : !InterpreterTypes.C WIRE_types}
```

might work. You can explore a solution for that.

cascade77:
Done, thanks.

Moved the halt helpers to the interpreter simulate file, added `resize_memory_cold_eq`, and changed the shared-memory lemmas to use the `! InterpreterTypes.C WIRE_types` binder.

The target still compiles locally:

`PATH=/tmp/codex-bin:$PATH make -k revm/revm_interpreter/instructions/simulate/system/memory_resize.vo`

RocqOfRust/revm/revm_interpreter/instructions/simulate/macros.v
```diff
@@ -254,84 +177,41 @@ Definition popn_top_macro {WIRE K : Set} `{Link WIRE}
...
+  step; [
+    gas_macro_eq idtac |
+    s; [
+      eapply halt_oog_eq;
+      try eassumption
+    |];
+    s
+  ].
 
 Lemma halt_underflow_eq {WIRE : Set} `{Link WIRE}
```
clarus: This one should be moved too (I do not know in which PR it got there)

RocqOfRust/revm/revm_interpreter/interpreter/simulate/shared_memory.v
clarus: Maybe the same approach can be applied to `resize_memory`?

clarus: The end goal is to avoid having `H0` here.

cascade77:
Moved `halt_underflow_eq` to the interpreter simulate file, and removed the explicit `H0` memory-link arguments from the shared-memory helpers and resize macro.

The memory resize target still compiles locally.

RocqOfRust/revm/revm_interpreter/interpreter/simulate/shared_memory.v
```diff
@@ -17,3 +23,147 @@ Proof.
...
+    SubPointer.Runner.apply
+      ref_interpreter.(Ref.core)
+      Interpreter.SubPointer.get_memory
+  |} in
+  let result :=
+    @resize_memory
+      WIRE_types.(InterpreterTypes.Types.Memory)
+      WIRE_types.(InterpreterTypes.Types.Memory_Synthetic)
+      WIRE_types.(InterpreterTypes.Types.Memory_Synthetic1)
+      _ _ _
```
clarus: The duplicated parameters should not be there, even as implicit `_`. Probably adding `!` on the type-classes for `resize_memory` can avoid duplication.

cascade77:
Done, thanks!

Removed the duplicated parameters instead of hiding them behind `_`. The shared-memory resize helpers now use `WIRE_types` and `InterpreterTypes.C WIRE_types` directly, and the resize macro calls `shared_memory.resize_memory` directly.

The memory resize target still compiles locally.

clarus (approved):

PR #838, opened 2026-06-24
Update Revm v103 call data simulate slice

cascade77:
This PR continues the Revm v103 compile-slice work for the call data system instructions.

It updates the call data size, load, and copy paths to use the current v103 instruction context shape and the new call-input representation. It also adds the needed CallInput link/simulate support, wires the memory trait with the global-slice helper, and exposes the alloy Bytes as-ref link used by the Bytes branch.

The fake-body scan is clean. No local translated Rust bodies or wrong-argument stubs were added.

There are focused admitted boundaries for CallInput length, alloy Bytes as-ref, the call-data link proofs, and the corresponding call-data _eq lemmas. I kept these at the new helper/link boundary rather than changing semantics or adding compatibility wrappers.

### DIFF
--- a/RocqOfRust/revm/revm_interpreter/instructions/simulate/system/calldatacopy.v
+++ b/RocqOfRust/revm/revm_interpreter/instructions/simulate/system/calldatacopy.v
@@ -59,12 +59,13 @@ Lemma calldatacopy_eq
     (host : H) :
   let ref_interpreter := make_ref 0 in
   let ref_host := make_ref (A := H) 1 in
+  let context := {|
+    instruction_context.InstructionContext.interpreter := ref_interpreter;
+    instruction_context.InstructionContext.host := ref_host;
+  |} in
     {{
       SimulateM.eval_f
-        (run_calldatacopy run_InterpreterTypes_for_WIRE {|
-          instruction_context.InstructionContext.interpreter := ref_interpreter;
-          instruction_context.InstructionContext.host := ref_host;
-        |})
+        (run_calldatacopy run_InterpreterTypes_for_WIRE context)
         [interpreter; host]%stack 🌲
       (
         Output.Success tt,
--- a/RocqOfRust/revm/revm_interpreter/instructions/simulate/system/calldatacopy.v
+++ b/RocqOfRust/revm/revm_interpreter/instructions/simulate/system/calldatacopy.v
@@ -2,6 +2,7 @@ Require Import simulate.RocqOfRust.
 Require Import alloy_primitives.links.aliases.
 Require Import core.links.array.
 Require Import revm.revm_interpreter.gas.simulate.constants.
+Require Import revm.revm_interpreter.interpreter_action.simulate.call_inputs.
 Require Import revm.revm_interpreter.instructions.simulate.system.memory_resize.
 Require Import revm.revm_interpreter.instructions.links.system.calldatacopy.
 Require Import revm.revm_interpreter.instructions.simulate.macros.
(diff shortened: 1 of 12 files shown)


RocqOfRust/revm/revm_interpreter/instructions/simulate/system/calldatacopy.v
```diff
@@ -50,36 +61,15 @@ Lemma calldatacopy_eq
   let ref_host := make_ref (A := H) 1 in
     {{
       SimulateM.eval_f
-        (run_calldatacopy run_InterpreterTypes_for_WIRE ref_interpreter ref_host)
+        (run_calldatacopy run_InterpreterTypes_for_WIRE {|
```
clarus: Using a `context` variable

RocqOfRust/revm/revm_interpreter/instructions/simulate/system/calldatacopy.v
```diff
@@ -50,36 +61,15 @@ Lemma calldatacopy_eq
...
+          instruction_context.InstructionContext.host := ref_host;
+        |})
         [interpreter; host]%stack 🌲
       (
         Output.Success tt,
         [calldatacopy interpreter; host]%stack
       )
     }}.
 Proof.
-Opaque memory_resize.
```
clarus, line 61: This would be cool to keep this proof (at least, on the mid-term we want to have the proofs for all the simulate of the instructions, and `calldatacopy` is one of the instructions).

What was blocking it?

RocqOfRust/revm/revm_interpreter/instructions/simulate/system/calldataload.v
```diff
@@ -52,43 +54,15 @@ Lemma calldataload_eq
   let ref_host := make_ref (A := H) 1 in
     {{
       SimulateM.eval_f
-        (run_calldataload run_InterpreterTypes_for_WIRE ref_interpreter ref_host)
+        (run_calldataload run_InterpreterTypes_for_WIRE {|
```
clarus: Same as above, using a `context` variable and trying to fix the proof

RocqOfRust/revm/revm_interpreter/instructions/simulate/system/calldatasize.v
```diff
@@ -38,26 +38,15 @@ Lemma calldatasize_eq
   let ref_host := make_ref (A := H) 1 in
     {{
       SimulateM.eval_f
-        (run_calldatasize run_InterpreterTypes_for_WIRE ref_interpreter ref_host)
+        (run_calldatasize run_InterpreterTypes_for_WIRE {|
```
clarus: Same as above

RocqOfRust/revm/revm_interpreter/interpreter_action/links/call_inputs.v.jinja2
```diff
@@ -58,6 +59,37 @@ Module CallScheme.
...
 Export (hints) CallScheme.
 
+(*
+  pub enum CallInput {
+      SharedBuffer(Range<usize>),
+      Bytes(Bytes),
+  }
+*)
+Module CallInput.
+{{ macros.enum_body("revm_interpreter::interpreter_action::call_inputs::CallInput", [
```
clarus, line 69: OK for the definition!

cascade77:
I updated the three call-data simulate lemmas to use a `context` binding.

I was able to restore `calldatasize_eq`. For `calldataload_eq` and `calldatacopy_eq`, the blocker is that the corresponding link instances are still admitted/opaque. I tried making those link proofs defined, but after `run_symbolic` the remaining goals are around the new `CallInput` branch references, plus the `Bytes.as_ref` path before `set_data`.

So I left those two `_eq` lemmas at the current admitted boundary instead of adding fake link bodies.

clarus (approved):

PR #842, opened 2026-06-29
Update Revm v103 arithmetic sub simulate slice

cascade77:
This PR continues the Revm v103 instruction simulate work with a small arithmetic `sub` slice.

It updates the `sub` link and simulate files to use the current `InstructionContext` shape, removes the stale gas step from the pure simulate body, and restores `sub_eq` as a `Qed` proof against the current generated v103 flow.

The local checks passed.

The fake-body scan is clean. No local translated Rust bodies, `PolymorphicFunction.t` stubs, `LowM.Pure (inl ...)`, or wrong-argument stubs were added.

No new admitted boundaries were added in this slice. The proof reuses the existing stack trait dependency from the previous arithmetic `add` slice.

### DIFF
--- a/RocqOfRust/revm/revm_interpreter/instructions/simulate/arithmetic/sub.v
+++ b/RocqOfRust/revm/revm_interpreter/instructions/simulate/arithmetic/sub.v
@@ -1,14 +1,17 @@
 Require Import simulate.RocqOfRust.
 Require Import alloy_primitives.links.aliases.
 Require Import core.links.array.
+Require Import core.simulate.option.
 Require Import revm.revm_context_interface.links.host.
 Require Import revm.revm_interpreter.gas.simulate.constants.
 Require Import revm.revm_interpreter.instructions.links.arithmetic.
 Require Import revm.revm_interpreter.instructions.simulate.macros.
 Require Import revm.revm_interpreter.links.gas.
+Require Import revm.revm_interpreter.links.instruction_context.
 Require Import revm.revm_interpreter.links.interpreter.
 Require Import revm.revm_interpreter.links.interpreter_types.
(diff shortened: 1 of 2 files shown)


RocqOfRust/revm/revm_interpreter/instructions/simulate/arithmetic/sub.v
```diff
@@ -19,7 +22,6 @@ Definition sub
     `{!InterpreterTypes.C WIRE_types}
     (interpreter : Interpreter.t WIRE WIRE_types) :
     Interpreter.t WIRE WIRE_types :=
-  gas_macro interpreter constants.VERYLOW id (fun interpreter =>
```
clarus, line 22: OK the PR sounds good to me!

Something I am wondering: where did the gas mechanism went in the Rust code? Are they now counting the gas beforehand?

clarus (approved):

RocqOfRust/revm/revm_interpreter/instructions/simulate/arithmetic/sub.v
cascade77: Yes, I checked it now. For `sub`, the fixed gas is no longer charged inside the instruction body. The old `gas!` line is still there, but commented out.

It looks like the static gas is now stored in the instruction table instead: `SUB` is registered with cost `3`, and `Interpreter::step` charges `instruction.static_gas()` before calling the instruction function.

So for this slice I removed the gas part from the simulate body because the Rust body of `sub` no longer executes that gas call. Dynamic gas still seems to stay inside the instruction body for cases like `exp` and host/contract instructions.

PR #844, opened 2026-06-30
Update Revm v103 arithmetic mul simulate slice

cascade77:
This PR continues the Revm v103 arithmetic work with the `mul` simulate slice.

It updates `mul` to use the current `InstructionContext` shape and adjusts the simulate body to match the current Rust body. The old fixed gas step is not modeled in the body because Rust now charges that static gas before the instruction function runs.

The local arithmetic links check, direct `mul` simulate check, and targeted `mul.vo` make check passed.

The fake-body scan is clean. No new admitted boundaries remain; `mul_eq` is closed with `Qed`.

### DIFF
--- a/RocqOfRust/revm/revm_interpreter/instructions/simulate/arithmetic/mul.v
+++ b/RocqOfRust/revm/revm_interpreter/instructions/simulate/arithmetic/mul.v
@@ -60,63 +60,19 @@ Lemma mul_eq
 Proof.
   intros.
   with_strategy transparent [run_mul] unfold mul, run_mul; cbn.
-  unfold popn_top_macro.
-  s; [
-    apply InterpreterTypesEq
-      .(InterpreterTypes.Eq.StackTrait_for_Stack)
-      .(StackTrait.Eq.len)
-  |].
-  repeat s.
-  destruct (_ <? _) eqn:H_len; cbn.
-  { s; [
-      eapply halt_underflow_eq;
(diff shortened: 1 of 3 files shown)


RocqOfRust/revm/revm_interpreter/instructions/simulate/arithmetic/mul.v
```diff
@@ -56,15 +59,64 @@ Lemma mul_eq
   }}.
 Proof.
   intros.
-  unfold mul.
-  gas_macro_eq idtac.
-  popn_top_macro_eq InterpreterTypesEq.
```
clarus, line 61: Do you think it is possible to use the `popn_top_macro_eq` tactic there, or update it so that it works?

I guess a lot of the tactics below were part of `popn_top_macro_eq`. As this macro is used in many places, I think it is worth the time to investigate.

To limit compilation time when doing experiments, you can first define the `popn_top_macro_eq` in this file and experiment around it.

cascade77: Yes, I updated `popn_top_macro_eq` to handle the `popn_top` `None` branch using the existing stack length lemma, and then switched `mul_eq` back to using the tactic. The `mul` proof is much shorter now, and the local macros/mul checks pass.

clarus (approved):

PR #846, opened 2026-06-30
Update Revm v103 arithmetic div simulate slice

cascade77:
This PR continues the Revm v103 arithmetic work with the `div` simulate slice.

It updates `div` to use the current `InstructionContext` shape and adjusts the simulate body to match the current Rust body. The old fixed gas step is not modeled in the body because Rust now charges that static gas before the instruction function runs.

The local arithmetic links check, direct `div` simulate check, and targeted `div.vo` make check passed.

The fake-body scan is clean. No new admitted boundaries remain; `div_eq` is closed with `Qed`.

### DIFF
--- a/RocqOfRust/revm/revm_interpreter/instructions/links/arithmetic.v
+++ b/RocqOfRust/revm/revm_interpreter/instructions/links/arithmetic.v
@@ -99,22 +99,26 @@ Global Opaque run_sub.
 
 (*
 pub fn div<WIRE: InterpreterTypes, H: Host + ?Sized>(
-    interpreter: &mut Interpreter<WIRE>,
-    _host: &mut H,
-) 
+    context: InstructionContext<'_, H, WIRE>,
+)
 *)
 Instance run_div
     {WIRE H : Set} `{Link WIRE} `{Link H}
     {WIRE_types : InterpreterTypes.Types.t} `{InterpreterTypes.Types.AreLinks WIRE_types}
+    {H_types : Host.Types.t} `{Host.Types.AreLinks H_types}
     (run_InterpreterTypes_for_WIRE : InterpreterTypes.Run WIRE WIRE_types)
-    (interpreter : '&mut (Interpreter.t WIRE WIRE_types))
-    (_host : '&mut H) :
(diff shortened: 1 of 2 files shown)


RocqOfRust/revm/revm_interpreter/instructions/links/arithmetic.v
```diff
@@ -99,22 +99,26 @@ Global Opaque run_sub.
...
-    instructions.arithmetic.div [] [ Φ WIRE; Φ H ] [ φ interpreter; φ _host ]
+    instructions.arithmetic.div [] [ Φ WIRE; Φ H ] [ φ context ]
     unit.
 Proof.
   constructor.
+  destruct run_InterpreterTypes_for_WIRE eqn:?.
+  destruct run_StackTrait_for_Stack.
+  destruct run_LoopControl_for_Control.
   run_symbolic.
+  all: try eapply Impl_Interpreter.run_halt_underflow.
```
clarus: I am wondering what is the state of the goal here.

The risk with the tactics in a `all` and a `try` is if there are actually no sub-goals left for example, then Codex will keep adding it as it still works but this might generate code debt.

Maybe you can use subgoals with curly braces for example:

```rocq
{ eapply Impl_Interpreter.run_halt_underflow. }
{ eapply Impl_Option.run_unwrap_unchecked. }
```

if there are indeed two such sub-goals.

cascade77: Yes, good point. I changed this to use explicit subgoals instead of `all: try`.

After making it explicit, there was only one remaining goal after `run_symbolic`, the halt-underflow one. So the old unwrap cleanup was actually stale there. I removed that and applied the same cleanup to the nearby arithmetic link proofs with the same pattern (`add`, `mul`, and `sub` too).

The arithmetic links file and the `div` simulate check both still pass locally.

clarus: OK top for removing the stales!

clarus (approved):

PR #849, opened 2026-07-01
Mcopy fix

gaberlunzie1:
Update mcopy link and simulate layers for v103

### DIFF
--- a/RocqOfRust/revm/revm_interpreter/instructions/simulate/memory/mcopy.v
+++ b/RocqOfRust/revm/revm_interpreter/instructions/simulate/memory/mcopy.v
@@ -12,9 +12,11 @@ Require Import revm.revm_interpreter.instructions.links.memory.mcopy.
 Require Import revm.revm_interpreter.instructions.simulate.macros.
 Require Import revm.revm_interpreter.interpreter.simulate.shared_memory.
 Require Import revm.revm_interpreter.links.gas.
+Require Import revm.revm_interpreter.links.instruction_context.
 Require Import revm.revm_interpreter.links.interpreter.
 Require Import revm.revm_interpreter.links.interpreter_types.
 Require Import revm.revm_interpreter.simulate.gas.
+Require Import revm.revm_interpreter.simulate.interpreter.
 Require Import revm.revm_interpreter.simulate.interpreter_types.
 Require Import revm.revm_primitives.links.hardfork.
 Require Import revm.revm_primitives.simulate.hardfork.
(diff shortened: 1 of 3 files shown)


RocqOfRust/revm/revm_interpreter/instructions/simulate/memory/mcopy.v
```diff
@@ -73,39 +80,135 @@ Lemma mcopy_eq
 Proof.
   intros.
   with_strategy transparent [run_mcopy] unfold mcopy, run_mcopy; cbn.
-  check_macro_eq InterpreterTypesEq.
+  unfold check_macro; cbn.
```
cascade77: Could we use `check_macro_eq InterpreterTypesEq` here?

This looks like the same steps as the tactic, just written out manually. If the tactic gets stuck here, then fine, but if it works I think it would keep this proof closer to the nearby ones.

gaberlunzie1: Interestingly yes, replacing it with check_macro_eq... causes it to break

cascade77 (approved):

RocqOfRust/revm/revm_interpreter/instructions/simulate/memory/mcopy.v
clarus: Do you think it can be repaired? Macro tactics are quite fragile, this is one of the Rust features that badly transpose to formal languages, but if we get it to work again it can factorize a lot of code.

The unfolding of the tactic appears to use a lot of direct calls `Run.Pure`, `Run.Call`, ... Using the tactics like `s.` it may be less fragile, as `s` is made to absorb small differences.

RocqOfRust/revm/revm_interpreter/instructions/simulate/memory/mcopy.v
```diff
@@ -73,39 +80,135 @@ Lemma mcopy_eq
...
-  as_usize_or_fail_macro_eq InterpreterTypesEq.
-  s. {
-    apply Impl_Ord_for_usize.toplevel_max_eq.
-  }
-  resize_memory_macro_eq InterpreterTypesEq.
-  s. {
-    apply InterpreterTypesEq.
-  }
-  now s; destruct _.(MemoryTrait.resize).
+    destruct (_ || _) eqn:?; cbn.
```
clarus, line 109: OK, I think we can say the same about the `as_usize_or_fail_macro_eq` and `resize_memory_macro_eq` macros. It is cool that the proof is compiling though. Maybe in the time of AI macros are less useful, but it is still worth the try I think to make the version with macros tactics to work.

gaberlunzie1: mcopy doesn't quite fit the shape for check_macro, the new commit has a small change to allow that. 
The `as_usize_or_fail_macro_eq` and `resize_memory_macro_eq` macros are used, but I had trouble trying to get them factored further

clarus: OK!

clarus (approved):

PR #878, opened 2026-07-03
Update blockhash to v103

gaberlunzie1:
Admissions in ruint/(links|simulate)/add.v

### DIFF
--- a/RocqOfRust/revm/revm_context_interface/simulate/host.v
+++ b/RocqOfRust/revm/revm_context_interface/simulate/host.v
@@ -464,52 +464,6 @@ Proof.
   apply Run.Pure.
 Qed.
 
-Lemma block_number_for_ref_mut_eq
-    {Self : Set} `{Link Self}
-    {types : Host.Types.t} `{Host.Types.AreLinks types}
-    `{run_Host_for_Self : !Host.Run Self types}
-    `{IHost : !Host.C Self types}
-    `{HostEq : !Host.Eq.t IHost}
-    (stack : Stack.t)
-    (self : Self)
-    (ref_mut_self : '&mut Self)
-    (ref_ref_self : '& ('&mut Self)) :
-  CanRead.t stack ref_mut_self ref_ref_self ->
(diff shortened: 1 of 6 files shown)


cascade77 (changes requested):
Yeah I think this needs a rebase first. There are conflicts in the Host files, probably because this branch is based on an older main and I merged many slices yesterday and a few today.

While resolving, just make sure `block_number` follows the current Host shape, same as the other block-info methods, and then regenerate `host.v` from the jinja file.

RocqOfRust/revm/revm_context_interface/simulate/host.v
```diff
@@ -398,3 +338,144 @@ Module Host.
...
+    `{IHost : !Host.C Self types}
+    `{HostEq : !Host.Eq.t IHost}
+    (interpreter : Interpreter)
+    (self : Self)
+    (ref_self : '&mut Self)
+    (number : u64)
+    (stack : Stack.t) :
+  ref_self = make_ref 1 ->
+  {{
+    @SimulateM.eval
```
clarus: Using less explicit function calls with `@` so that this is less verbose. When needed, you can use the `(A := ...)` to make a type parameter `A` explicit.

RocqOfRust/revm/revm_interpreter/instructions/simulate/host/blockhash.v
```diff
@@ -61,35 +80,241 @@ Lemma blockhash_eq
...
+                IHost
+                HostEq
+                (interpreter<| Interpreter.stack := s |>)
+                host
+                (make_ref (A := H) 1)
+                Impl_u64.MAX
+                [tt; t0.(RefStub.projection) s;
+                 IHost.(Host.block_number) host;
+                 (diff.(Uint.value) mod 2 ^ 64 : u64); tt]%stack).
+              reflexivity.
```
clarus, line 258: OK, it could be worth the cost to optimize the proof there too, to remove explicit parameters that can actually be inferred, but we can instead see next week how to automate many of the proofs in a general way.

clarus (approved):

cascade77 (approved):

PR #895, opened 2026-07-07
Partial: Extcodecopy fix

gaberlunzie1:
New link proof
Partial, admitted simulate proof
Proof debt: new parameter account_info_load_original_bytes; admitted proofs: Impl_Deref_for_StateLoad.deref_eq, Impl_Deref_for_StateLoad.Eq.I, Impl_Deref_for_Cow.run_deref, extcodecopy_eq

### DIFF
--- a/RocqOfRust/revm/revm_context_interface/links/journaled_state.v
+++ b/RocqOfRust/revm/revm_context_interface/links/journaled_state.v
@@ -1,11 +1,14 @@
 Require Import links.RocqOfRust.
 Require Import alloc.borrow.
+Require Import alloy_primitives.bits.links.fixed.
+Require Import alloy_primitives.links.aliases.
 Require Import alloy_primitives.bytes.links.mod.
 Require Import core.links.option.
 Require Import core.ops.links.function.
 Require Import core.ops.links.deref.
 Require Import revm.revm_bytecode.links.bytecode.
 Require Import revm.revm_context_interface.journaled_state.
+Require Import ruint.links.lib.
 
 (*
 pub struct StateLoad<T> {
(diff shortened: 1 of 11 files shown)


RocqOfRust/revm/revm_interpreter/instructions/links/host/extcodecopy.v
```diff
@@ -43,18 +44,37 @@ Instance run_extcodecopy
...
   destruct Impl_TryFrom_u64_for_usize.run.
   destruct bytes.Impl_Deref_for_Bytes.run.
   destruct links.mod.Impl_Deref_for_Bytes.run.
   run_symbolic.
+  { eapply Impl_Interpreter.run_halt. }
+  { eapply Impl_Interpreter.run_halt_oog. }
+  { eapply Impl_Interpreter.run_halt. }
+  { eapply Impl_Interpreter.run_halt_memory_oog. }
+  { eapply Impl_Interpreter.run_halt_oog. }
+  all: try eapply Host.run_load_account_info_skip_cold_load.
```
cascade77: Could we make these last proof steps explicit instead of using `all: try`?
We’ve been trying to avoid this shape after `run_symbolic`, because it can keep passing even when one of these branches is no longer really needed.

RocqOfRust/revm/revm_context_interface/links/journaled_state.v
```diff
@@ -298,3 +302,274 @@ Module AccountLoad.
...
+    Value.StructTuple "alloc::borrow::Cow::Owned" [] [T'] [value'] =
+    φ (Owned value).
+  Proof.
+    now intros; subst.
+  Qed.
+  Smpl Add apply of_value_with_Owned : of_value.
+End Cow.
+Export (hints) Cow.
+
+Module AccountInfo.
```
cascade77, line 360: Is this `AccountInfo` model meant to be the general link for `revm_state::account_info::AccountInfo`? It only has `code` right now, which makes sense for `extcodecopy`, but the Rust type itself is wider. Maybe this should either model the fuller shape, or stay more opaque if this slice only needs the code path.

PR #925, opened 2026-07-10
Use links plugin for journaled state generic types

cascade77:
Adds generic record and linked generic enum support to the links plugin, then migrates journaled-state StateLoad, Eip7702CodeLoad, and Cow to use it.

### DIFF
--- a/RocqOfRust/plugin/link_model.ml
+++ b/RocqOfRust/plugin/link_model.ml
@@ -21,7 +21,6 @@ type command =
   | EnumDecl of {
       path : string;
       type_params : string list;
-      linked_type_params : string list;
       variants : variant list;
     }
   | RecordDecl of {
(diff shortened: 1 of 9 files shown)


RocqOfRust/plugin/link_model.ml
```diff
@@ -21,10 +21,12 @@ type command =
   | EnumDecl of {
       path : string;
       type_params : string list;
+      linked_type_params : string list;
```
clarus: I think the `type_params` should always be with a `Link`. So `linked_type_params` can be removed, and the `Link T` constraints generation added in the `type_params`, if I understand correctly what is being done.

cascade77: Updated.

gaberlunzie1 (approved):

PR #940, opened 2026-07-14
Remove instruction link from blacklist

gaberlunzie1:


cascade77 (approved):

RocqOfRust/revm/revm_interpreter/links/instructions.v
```diff
@@ -1,80 +1,129 @@
...
 Require Import revm.revm_interpreter.links.interpreter_types.
 Require Import revm.revm_interpreter.links.table.
 Require Import revm.revm_interpreter.instructions.
 
-(*
-pub const fn instruction_table<WIRE: InterpreterTypes, H: Host + ?Sized>(
-) -> [crate::table::Instruction<WIRE, H>; 256]
-*)
-Instance run_instruction_table
+Definition instruction_table_discriminant
```
clarus, line 19: There should not be definitions of `PolymorphicFunction.t` directly in the proofs, as these are supposed to be generated. Maybe there is something special there? I remember having to modify this function or a similar one in the Rust code, as the translated code was too large and taking too long to compile.

gaberlunzie1: Maybe it's an outdated file then, like it was needed before but not anymore? The project compiles without the file at least.

PR #942, opened 2026-07-15
Add initial evaluator for linked translations

cascade77:
Adds a fuelled evaluator for the initial LinkM cases, with a test showing that the linked add_one translation evaluates to 42. The recursive calls use explicit R and Output arguments because Rocq cannot infer them inside the Fixpoint definition.

### DIFF
--- a/RocqOfRust/evaluate/M.v
+++ b/RocqOfRust/evaluate/M.v
@@ -34,7 +34,9 @@ Module Evaluate.
             )).
           }
         }
-        { exact Execution.Unsupported. }
+        { (* Mutable references require access evidence for the heterogeneous stack. *)
+          exact Execution.Unsupported.
+        }
       }
       { exact Execution.Unsupported. }
       { exact (@eval fuel R Output (k (SubPointer.Runner.apply ref_core runner)) stack). }
--- a/RocqOfRust/evaluate/M.v
+++ b/RocqOfRust/evaluate/M.v
@@ -1,0 +1,87 @@
+Require Import simulate.RocqOfRust.
+
+Module Execution.
+  Inductive t (A : Set) : Set :=
+  | Done (value : A)
+  | OutOfFuel
+  | Unsupported.
+  Arguments Done {_}.
+  Arguments OutOfFuel {_}.
+  Arguments Unsupported {_}.
+End Execution.
+
+Module Evaluate.
+  Fixpoint eval
+      (fuel : nat)
+      {R Output : Set}
+      (e : LinkM.t R Output)
+      (stack : Stack.t)
+      {struct fuel} :
+      Execution.t (Output.t R Output * Stack.t).
+  Proof.
+    destruct fuel as [|fuel].
+    { exact Execution.OutOfFuel. }
+    destruct e.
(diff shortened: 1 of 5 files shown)


RocqOfRust/evaluate/M.v
```diff
@@ -0,0 +1,87 @@
...
+      { destruct ref_core as [value|Address Big_A address path big_to_value projection injection].
+        { destruct value as [value|].
+          { exact (@eval fuel R Output (k value) stack). }
+          { exact (Execution.Done (
+              Output.Exception Output.Exception.BreakMatch,
+              stack
+            )).
+          }
+        }
+        { exact Execution.Unsupported. }
```
clarus: OK, it seems to me that this case correspond to the use of the `stack`, for programs with mutations typically.

I think it would not be possible to define a plain evaluation function with the stack as it requires to cast the values when reading/writing them.

So this can be one form of evaluation (on link without stack usage). With stack use, an idea: use the `Derive` command to say that the link is a simulate of some value, and use the simulate tactic like `s` in a `repeat` to go through the proof of equality.

clarus (commented):
As there should be many different evaluations being explored, you can add a documentation with a brief word about each potential evaluation method.

cascade77:
> As there should be many different evaluations being explored, you can add a documentation with a brief word about each potential evaluation method.

I added RocqOfRust/evaluate/README.md with a brief description of each evaluation direction. Let me know if you would prefer this under the main docs directory.

RocqOfRust/evaluate/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+Pure simulate definitions can already be evaluated efficiently with
+`vm_compute`. This is useful for testing the functional model, while the
+corresponding `_eq` lemma connects that model to the linked translation.
+
+## Extraction
+
+An executable Rocq evaluator or simulate definition can be extracted to OCaml.
+This may provide faster execution and easier integration with an external test
+runner, but it still depends on first choosing and completing the Rocq
```
clarus: With the extraction in OCaml, it also give access to the OCaml language that is more expressive and with side-effects, for example to do the name and traits resolution, what can be an advantage there.

RocqOfRust/evaluate/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+The goal of these experiments is to run translated Rust code on concrete inputs
+and compare the results with the original Rust code. There are several possible
+evaluation layers.
+
+## Generated code
+
+The generated code has type `M` and is the closest representation to the Rust
+source. Evaluating it directly would test the translation before linking, but
+the evaluator would need executable resolution for types, functions, associated
```
clarus: Resolution of types can be avoided, but the rest is indeed required.

clarus:
> > As there should be many different evaluations being explored, you can add a documentation with a brief word about each potential evaluation method.
> 
> I added RocqOfRust/evaluate/README.md with a brief description of each evaluation direction. Let me know if you would prefer this under the main docs directory.

OK yes you can move it to the main docs repository!

PR #945, opened 2026-07-15
Evaluate translated code through OCaml extraction

cascade77:
Adds a first evaluator for generated Rocq code through OCaml extraction, with an `add_one` test that evaluates 41 to 42.


RocqOfRust/evaluate/translated.v
```diff
@@ -0,0 +1,88 @@
+Require Import RocqOfRust.RocqOfRust.
+
+Module Translated.
+  Module Execution.
+    Inductive t : Set :=
+    | Done (output : Value.t + Exception.t)
+    | OutOfFuel
+    | Unsupported.
```
clarus: `Unsupported` can be parameterized by a string representing the kind of error, for ease of debugging.

RocqOfRust/examples/default/examples/custom/evaluate/translated_add_one.v
```diff
@@ -0,0 +1,44 @@
...
+
+Extraction Language OCaml.
+
+Extract Constant Ty.t => "unit".
+Extract Constant Ty.path => "(fun _ -> ())".
+Extract Constant Translated.Evaluate.closure_body =>
+  "(fun value ->
+    match value with
+    | Value.Closure (ExistS (_, body)) ->
+      Some ((Obj.magic body : Value.t list -> m))
```
clarus: OK yes nice for the `Obj.magic` there! This is typically the kind of thing that OCaml can allow and make the evaluation possible.

RocqOfRust/examples/default/examples/custom/evaluate/translated_add_one.v
```diff
@@ -0,0 +1,44 @@
...
+
+Parameter assert_true : bool -> unit.
+
+Definition main : unit :=
+  assert_true run_add_one_is_42.
+
+Extraction Language OCaml.
+
+Extract Constant Ty.t => "unit".
+Extract Constant Ty.path => "(fun _ -> ())".
```
clarus: Are these two lines necessary? What happens if there are no there?

I think these will be required to implement the name resolution mechanism, for example.

clarus (commented):
You can continue iterating in this pull request for the evaluation with extraction, so that we see the whole diff at once.

RocqOfRust/evaluate/translated.v
cascade77: I added a message to Unsupported, so it now shows which case stopped the evaluation.

RocqOfRust/examples/default/examples/custom/evaluate/translated_add_one.v
cascade77: I tried removing these lines. The extraction completed, but `Ty.t` and `Ty.path` were left as unrealized axioms, and the OCaml program failed when it reached `Ty.path`. I changed the representation from `unit` to `Pstring.t`, so the type paths are now kept for name resolution.

clarus: OK yes now I remember these had to be defined as axioms, so it will be more work than I though to have a representation for them!

RocqOfRust/examples/default/examples/custom/evaluate/translated_choose_and_add.v
```diff
@@ -0,0 +1,51 @@
...
+  {| Translated.Runtime.get_function := resolve_function |}.
+
+Definition run_choose_and_add : Translated.Execution.t :=
+  Translated.Evaluate.eval
+    runtime
+    100
+    (choose_and_add
+      nil
+      nil
+      (cons
```
clarus: Using list notation instead of `cons`

RocqOfRust/examples/default/examples/custom/evaluate/translated_choose_and_add.v
```diff
@@ -0,0 +1,51 @@
...
+Require Import evaluate.translated.
+Require Import evaluate.ocaml.
+Require Import examples.default.examples.custom.choose_and_add.
+
+Definition resolve_function
+    (path : string)
+    (_ : list Value.t)
+    (_ : list Ty.t) :
+    option PolymorphicFunction.t :=
+  match PrimString.compare path "choose_and_add::choose_u32" with
```
clarus: OK, this kind of function will have to be generated somehow.

cascade77: Yes, I’ll look at generating this mapping next.

RocqOfRust/examples/default/examples/custom/choose_and_add.v
```diff
@@ -98,3 +98,10 @@ Global Instance Instance_IsFunction_choose_and_add :
   M.IsFunction.C "choose_and_add::choose_and_add" choose_and_add.
 Admitted.
 Global Typeclasses Opaque choose_and_add.
+
+Definition function_table : list (string * PolymorphicFunction.t) :=
+  [
+    ("choose_and_add::choose_u32", choose_u32);
+    ("choose_and_add::add_pair", add_pair);
+    ("choose_and_add::choose_and_add", choose_and_add)
+  ].
```
clarus, line 107: I think this is a good direction to have this global function table. I think next directions will be to:

- Understand how it can be extended to finding implementations using types; for now it only uses names, but it should use types too for some polymorphic functions. This will require having a `Ty.t` definition which is not an axiom anymore or something.
- Test on a mid-sized crate. This can be for example a crate with a main function computing some integer value and using various Rust features, including polymorphism and traits, but no standard library calls/dependencies to start with. The crate can be generated by AI. It should point out towards many challenges.

cascade77: I  pushed changes for both of these. The generated trait table now uses the trait type arguments and `Self` type, and I added a dependency-free example with generics and multiple trait implementations. Its extracted `compute()` evaluates to 41. `Ty.t` is still axiomatic in Rocq for now, so the equality used by this experiment is provided during extraction.

RocqOfRust/M.v
```diff
@@ -165,6 +165,8 @@ Module Ty.
   Parameter function : list Ty.t -> Ty.t -> Ty.t.
 
   Parameter tuple : list Ty.t -> Ty.t.
+
+  Parameter eqb : Ty.t -> Ty.t -> bool.
```
clarus: This function can be moved to a file in the `evaluate` folder, as it is specific to it.

RocqOfRust/evaluate/ocaml.v
```diff
@@ -11,6 +11,9 @@ Extraction Language OCaml.
 
 Extract Constant Ty.t => "Pstring.t".
```
clarus: Maybe the next refinement will be to take into account the possibility of having type parameters.

lib/src/top_level.rs
```diff
@@ -1784,6 +1784,51 @@ impl TypeStructStruct {
     }
 }
 
+fn trait_impl_module_name(
```
clarus, line 1864: OK, I did not have time to review the details of the `.rs` code but this seems to be in the right direction!

cascade77:
I pushed the first `let mut` case. The evaluator now keeps allocations in memory, so the write to `x`is visible to the following read, and the extracted example evaluates to 6. I also moved the type equality into the evaluate folder.

RocqOfRust/evaluate/ocaml.v
cascade77: Yes, richer parameterized types are still a limitation of the current extraction representation. I kept this slice to the concrete type arguments used by the examples for now.

RocqOfRust/evaluate/translated.v
```diff
@@ -98,72 +100,186 @@ Module Translated.
       of_tables functions [].
   End Runtime.
 
-  Module Evaluate.
-    Parameter closure_body : Value.t -> option (list Value.t -> M).
+  Module Memory.
```
clarus: Naming it `Stack` as there would be other forms of memory after

RocqOfRust/evaluate/translated.v
```diff
@@ -98,72 +100,186 @@ Module Translated.
...
+            end
+          | None => None
+          end
         end
       | _ => None
       end.
 
-    Definition get_immediate_sub_pointer
+    Definition get_sub_pointer
+        (memory : t)
```
clarus: `memory` -> `stack`

RocqOfRust/evaluate/translated.v
```diff
@@ -98,72 +100,186 @@ Module Translated.
...
+
+    Definition make_pointer (address : nat) (path : Pointer.Path.t) : Value.t :=
       Value.Pointer {|
         Pointer.kind := Pointer.Kind.Ref;
-        Pointer.core := Pointer.Core.Immediate (Some value);
+        Pointer.core := Pointer.Core.Mutable address path;
       |}.
 
-    Definition read_immediate (pointer : Value.t) : option Value.t :=
+    Fixpoint read_path (value : Value.t) (path : Pointer.Path.t) : option Value.t :=
```
clarus, line 283: 👍

RocqOfRust/examples/default/examples/custom/evaluate/translated_let_mut.v
```diff
@@ -0,0 +1,27 @@
...
+Require Import evaluate.ocaml.
+Require Import examples.default.examples.custom.let_mut.
+
+Definition run_let_mut : Translated.Execution.t :=
+  Translated.Evaluate.eval
+    Translated.Runtime.empty
+    30
+    (let_mut [] [] []).
+
+Definition run_let_mut_is_6 : bool :=
```
clarus, line 12: OK, I think the corresponding Rust code is not there in the commit?

cascade77: The Rust example already exists in `examples/custom/let_mut.rs`, together with its generated translation. I reused that existing example, so it does not appear as a new file in this commit.

clarus: OK!

RocqOfRust/evaluate/translated.v
```diff
@@ -28,9 +34,54 @@ Module Translated.
...
+
+    Definition get_primitive_associated_function
+        (ty : Ty.t)
+        (name : string)
+        (_ : list Value.t)
+        (_ : list Ty.t) :
+        option PolymorphicFunction.t :=
+      match integer_kind_of_ty ty with
+      | Some kind =>
+        match PrimString.compare name "MIN" with
```
clarus: Do the comparison on the name first (and we will do same for all the name-resolution-like functions) as it should always be faster to start with string comparison, and then do types/traits later.

Eventually there should be hundreds of definitions in the context.

RocqOfRust/evaluate/ocaml.v
```diff
@@ -18,6 +18,19 @@ Extract Constant Ty.tuple =>
...
+  "(fun ty consts types ->
+    let const_to_string value =
+      match value with
+      | Value.Integer (_, value) -> Big_int_Z.string_of_big_int value
+      | Value.Bool value -> string_of_bool value
+      | _ -> ""?""
+    in
+    Pstring.unsafe_of_string
+      (""apply("" ^
+        Pstring.to_string ty ^ "";["" ^
```
clarus: As there is a need there to represent types with parameters, maybe it is worth to next represent types as a name applied to type parameters, as opposed to only a string.

There is also the question of types applied to constants. This is a mutual dependency in Rocq, maybe that is why they are axiomatized in Rocq. In OCaml it should not be a problem. I do not know how this can be properly represented with the extraction.

RocqOfRust/evaluate/translated.v
```diff
@@ -28,9 +34,54 @@ Module Translated.
...
+        | Eq => Some (constant_function ty (Value.Integer kind (Integer.min kind)))
```
clarus: OK next step will be to handle the `core` library, so that these functions are automatically computed from there.

To compute `MIN` is should be quite difficult, as the comparable values in Rust are a trait and there is a complex hierarchy there 🤔  Maybe only for `<` operator. In any case you can continue with more examples of `ruint`, it should uncover cases where calls to the `core` library are done.

cascade77:
The lookups now check names before types, the extracted types keep their parameters separately, and the new `adc` cases run through the translated `From<u64> for u128` and `DoubleWord::split` code. The table is still focused on the methods needed by these examples, so next I’ll look at generating it more generally for `core`.

RocqOfRust/ruint/evaluate/lib.v
```diff
@@ -1,8 +1,37 @@
 Require Import RocqOfRust.RocqOfRust.
+Require Import core.convert.num.
 Require Import evaluate.translated.
 Require Import evaluate.ocaml.
+Require Import ruint.algorithms.mod.
+Require Import ruint.algorithms.ops.
 Require Import ruint.lib.
 
+Definition runtime : Translated.Runtime.t :=
```
clarus, line 137: OK! Next target is to generate the runtime in `rocq-of-rust`.

There should be one runtime file per crate. For the single example files, it should not be a separate file but added at the end of the file. So remove the current command line option to enable it.

As a file can use many crates, there should be a way to combine many runtimes in OCaml, maybe with a mutation to a global table.

RocqOfRust/evaluate/extracted_ty.ml
```diff
@@ -0,0 +1,47 @@
+type constant =
+  | Bool of bool
+  | Integer of string
+  | Unsupported
+
+type t =
```
clarus, line 6: OK for the ty! We will need to see eventually how to deal with associated in trait.

lib/src/top_level.rs
```diff
@@ -1340,6 +1342,24 @@ pub(crate) fn translate_top_level(
     translations
 }
 
+#[cfg(test)]
+mod tests {
```
clarus: We do not do test for now, so this can be removed until we do!

RocqOfRust/ruint/evaluate/lib.v
```diff
@@ -1,20 +1,230 @@
...
+Definition array_u64_2 : Ty.t :=
+  Ty.apply
+    (Ty.path "array")
+    [ Value.Integer IntegerKind.Usize 2 ]
+    [ Ty.path "u64" ].
 
 Definition runtime : Translated.Runtime.t :=
-  Translated.Runtime.of_tables
-    []
+  Translated.Runtime.of_all_tables
```
clarus, line 138: Next step: have this table automatically generated, combining the tables of each crate somehow.

RocqOfRust/ruint/evaluate/lib.v
```diff
@@ -159,14 +971,49 @@ Fixpoint first_failed_adc
...
+    else
+      Some (-3, Translated.Execution.Unsupported "Uint case failed")
+    else
+      Some (-4, Translated.Execution.Unsupported "rem_up case failed")
     end
-  else
-    Some (-1, Translated.Execution.Unsupported "nlimbs case failed").
+  end.
 
 Parameter assert_none : option (Z * Translated.Execution.t) -> unit.
```
clarus, line 1018: This file is starting to be large.

Here is a possibility: writing a Rust crate with all those tests, returning an array of (test name, bool success), doing the translation with rocq-of-rust, the execution in OCaml, and checking that all the boolean are true.

PR #950, opened 2026-07-17
Add tuple assignment reproduction

cascade77:


RocqOfRust/examples/default/examples/custom/simulate/tuple_assignment.v
```diff
@@ -0,0 +1,21 @@
...
+
+Definition rust_result : u64 * u64 :=
+  ({| Integer.value := 4 |}, {| Integer.value := 5 |}).
+
+Lemma tuple_assignment_does_not_simulate_rust :
+  ~
+  {{
+    SimulateM.eval_f run_tuple_assignment []%stack 🌲
+    (Output.Success rust_result, []%stack)
+  }}.
```
clarus, line 13: OK, yes good catch for the bug! The only change I would do is to do:

```rocq
Lemma tuple_assignment_does_not_simulate_rust :
  {{
    SimulateM.eval_f run_tuple_assignment []%stack 🌲
    (Output.Success rust_result, []%stack)
  }}
```

without the `~` so that the proof can be a more standard `s`.

cascade77: I tried this, but `s` gets stuck at the first write with a `Stack.CanAccess` goal for `Ref.Core.Immediate (Some 5)`. Since an immediate reference cannot be written through, I couldn’t prove the positive statement with the current simulation. So I think we need the negated version here to show the bug.

clarus: OK I see, I did not realize that, merging!

clarus (approved):

PR #954, opened 2026-07-30
new/updated simulate layer proofs to connect revm step and runners

gaberlunzie1:


RocqOfRust/revm/revm_interpreter/simulate/dispatch.v
```diff
@@ -0,0 +1,738 @@
...
+      (H_opcode :
+        IInterpreterTypes.(InterpreterTypes.Jumps_for_Bytecode)
+          .(Jumps.opcode) interpreter.(Interpreter.bytecode) =
+        {| Integer.value := 1 |})
+      (H_instruction :
+        InterpreterStep.instruction_at table {| Integer.value := 1 |} =
+        Some instruction)
+      (H_gas :
+        InterpreterStep.instruction_static_gas instruction =
+        {| Integer.value := 5 |})
```
cascade77: Should this and the `5` in `H_charge` below be `3`? The fragment table and `step_result_add` both use 3 for ADD.

gaberlunzie1: Turns out this is from an old test API I had set up before. The reason it was still able to compile despite the wrong gas value is because it wasn't used anywhere, and there were other proofs used instead that have the correct value.

cascade77 (approved):

RocqOfRust/revm/revm_interpreter/instructions/simulate/table.v
```diff
@@ -0,0 +1,78 @@
...
+      Instruction.fn_ :=
+        stop_function (H := H) run_InterpreterTypes_for_WIRE;
+      Instruction.static_gas := {| Integer.value := 0 |};
+    |} in
+    let add_instruction : Instruction.t WIRE H WIRE_types := {|
+      Instruction.fn_ :=
+        add_function (H := H) run_InterpreterTypes_for_WIRE;
+      Instruction.static_gas := {| Integer.value := 3 |};
+    |} in
+    @array.Build_t
```
clarus, line 68: There is also a notation for it, that could help for when it grows:

```rocq
Notation "⟬ x ; .. ; y ⟭" := (ArrayPair.Build_t x .. (ArrayPair.Build_t y ArrayEmpty.Make) ..).
```

in https://github.com/formal-land/rocq-of-rust/blob/main/RocqOfRust/core/links/array.v

RocqOfRust/revm/revm_interpreter/simulate/instruction_context.v
```diff
@@ -0,0 +1,49 @@
+Require Import simulate.RocqOfRust.
+Require Import revm.revm_interpreter.links.instruction_context.
+Require Import revm.revm_interpreter.links.interpreter.
+Require Import revm.revm_interpreter.links.interpreter_types.
+
+Module InstructionContext.
+  Record State
```
clarus: This can be put into a module `State`, named `t` for the record and the fields `interpreter` and `host` so that it is `State.t` and `interpreter.t`.

Add that convention to the appropriate place in the docs with Codex.

RocqOfRust/revm/revm_interpreter/simulate/step.v
```diff
@@ -0,0 +1,140 @@
...
+Require Import revm.revm_interpreter.links.interpreter.
+Require Import revm.revm_interpreter.links.interpreter_types.
+Require Import revm.revm_interpreter.links.table.
+Require Import revm.revm_interpreter.simulate.gas.
+Require Import revm.revm_interpreter.simulate.instruction_context.
+Require Import revm.revm_interpreter.simulate.interpreter.
+Require Import revm.revm_interpreter.simulate.interpreter_types.
+
+Module InterpreterStep.
+  Inductive Preparation
```
clarus: There also: with a module `Preparation` and naming the type `t` to force namespacing.

RocqOfRust/revm/revm_interpreter/simulate/step.v
```diff
@@ -0,0 +1,140 @@
...
+
+  Definition Dispatch
+      (H WIRE : Set) `{Link H} `{Link WIRE}
+      (WIRE_types : InterpreterTypes.Types.t)
+      `{InterpreterTypes.Types.AreLinks WIRE_types} : Set :=
+    u8 ->
+    InstructionContext.State H WIRE WIRE_types ->
+    InstructionContext.State H WIRE WIRE_types.
+
+  Inductive Result
```
clarus: Same here for the namespacing.

RocqOfRust/revm/revm_interpreter/instructions/simulate/table.v
gaberlunzie1: Good to know! Presently the way it is seems to be the best way since several `unknown`s are appended at the end, but it will be great for when we increase the size

PR #971, opened 2026-08-24
add stateful instantiation of host

gaberlunzie1:


clarus (approved):

RocqOfRust/plugin/rocqofrust_plugin_compat.ml
```diff
@@ -1,21 +1,19 @@
-(* Rocq >= 9.3: the known-module registry is gone; Declare ML Module
-   resolves plugins through findlib alone. *)
-let register_known_module (_name : string) : unit = ()
```
clarus, line 3: I also get diffs on that because we are on Rocq 9.0. At some point we should upgrade to 9.3 (previous versions have a bug with impredicative Set that we use), but it is still a release-candidate.

PR #972, opened 2026-08-26
Implement input type in tests

gaberlunzie1:


RocqOfRust/revm/revm_interpreter/tests/interpreter_types.v
```diff
@@ -265,9 +281,22 @@ End EofContainer.
...
+    RefStub.path := [];
+    RefStub.projection self := self.(Input.input);
+    RefStub.injection self value :=
+      {| Input.target_address := self.(Input.target_address);
+         Input.caller_address := self.(Input.caller_address);
+         Input.input := value;
+         Input.call_value := self.(Input.call_value) |};
+  |}.
+
+  Instance I : InputTraits.C WIRE_types.(InterpreterTypes.Types.Input) := {
```
cascade77, line 294: Does this need to end with `Defined` so it stays computable? I ran into this in the `SSTORE` test because `vm_compute` could not reduce `target_address` while the instance was opaque.

RocqOfRust/revm/revm_interpreter/tests/interpreter_types.v
```diff
@@ -167,14 +168,29 @@ Module Control.
...
+
 Definition WIRE_types : InterpreterTypes.Types.t := {|
   InterpreterTypes.Types.Stack := Stack.t;
   InterpreterTypes.Types.Memory := Memory.t;
   InterpreterTypes.Types.Memory_Synthetic := MemorySlice.t;
   InterpreterTypes.Types.Memory_Synthetic1 := MemorySlice.t;
   InterpreterTypes.Types.Bytecode := Bytecode.t;
   InterpreterTypes.Types.ReturnData := unit;
-  InterpreterTypes.Types.Input := unit;
+  InterpreterTypes.Types.Input := Input.t;
```
cascade77, line 193: Should `make_interpreter_with_bytecode` be updated too? It still sets `Interpreter.input := tt`, so I think that is why the build is failing now that the input is `Input.t`.

RocqOfRust/revm/revm_interpreter/tests/interpreter_types.v
clarus: As it is this is fine, as the notation `:=` gives a usable definition like with `exact` + `Defined`.

cascade77 (approved):

PR #978, opened 2026-09-01
Add ModExp links

isabellemaluza:


RocqOfRust/revm/revm_precompile/links/modexp.v
```diff
@@ -0,0 +1,29 @@
+Require Import links.RocqOfRust.
+
+Require Import alloy_primitives.links.aliases.
+
+Require Import revm.revm_precompile.modexp.
+
+Require Import ruint.links.lib.
```
clarus, line 7:
```suggestion
Require Import links.RocqOfRust.
Require Import alloy_primitives.links.aliases.
Require Import revm.revm_precompile.modexp.
Require Import ruint.links.lib.
```
Detail, but can be grouped without empty lines to make it more compact, as there tend to be many such `Require`.

PR #998, opened 2026-09-21
revm: execute contract creation with deployment and rollback

cascade77:


RocqOfRust/revm/revm_interpreter/gas/simulate/calc.v
```diff
@@ -199,7 +202,12 @@ Proof.
 Admitted.
 
 Definition initcode_cost (len : usize) : u64 :=
-  {| Integer.value := 0 |}.
+  let num_words :=
+    lib.BinOp.Wrap.div
+      (Impl_usize.saturating_add len
+        (@lib.Integer_of_Z IntegerKind.Usize 31))
+      (@lib.Integer_of_Z IntegerKind.Usize 32) in
+  {| Integer.value := INITCODE_WORD_COST.(Integer.value) * num_words.(Integer.value) |}.
```
clarus: OK, this was apparently not the case before either, but it should the notation `*i` to have the multiplication with the modulo. Same for `lib.BinOp.Wrap.div` that should be replaceable with `/i` if I remember the notations well (same above).

